PR #45, opened 2015-01-07
Improve frontend private asset handling

anarchivist:
* Define new `frontend_use_restricted_assets` variable (boolean).
    * If True, copy private key; fetch private asset gem; build as normal; delete key.
    * If False, just build the frontend without the private asset gem.
* See dpla/frontend#45 for related frontend changes.


anarchivist:
OK - ready to merge.

PR #48, opened 2015-01-09
Ensure frontend uploads are owned by dpla:dpla [fixes #7769]

anarchivist:


markbreedlove:
That would do it; thanks for the catch!
I will merge this.

PR #50, opened 2015-01-09
Set up S3 configuration variables for frontend [refs #7734]

anarchivist:
See also dpla/frontend#46


markbreedlove:
This is the referenced issue:  https://issues.dp.la/issues/7734

Per personal conversations, this is good to merge.

PR #51, opened 2015-01-15
Add ingestion2

markbreedlove:
This changeset introduces new services and hosts for ingestion2 (Krikri, Heidrun, and their stack).

For development, there is a new inventory, a new Vagrantfile, and a new playbook to provide a two-VM environment that groups the services differently to save memory.  When we go live and stop using BigCouch, this will be revised to use more smaller VMs again, and the extra inventory and playbook file will be removed.  With the new set of files you can go back and forth between pre- and post-ingestion2 setups by running different sets of VMs.

New roles have been added:
- redis
- tomcat_common
- marmotta
- solr
- worker
- ingestion_app

Some things might be slightly controversial, so be sure to consider them:
- There are alternate uses of the terms "ingestion" and "heidrun" across various files, roles, and so on.  I try to use "heidrun" when the task can not have to do with anything other than the Heidrun code itself.  I have tried to use "ingestion" where it feels more general.  Still, these differences could be confusing.
- `automation` is maintaining its own Solr `schema.xml` file, similarly to how an automation system might assume responsibility for other files, like database configurations or frontend / api / wordpress configurations.  (We also create roles and permissions for application databases -- the list goes on.)  There is a `schema.xml` file maintained in the Krikri repository, but the deployment of Heidrun specifies a particular version of Heidrun, which, in turn, has a gemspec that requires a specific version of Krikri, and grabbing the right version of the schema file and then delegating its installation to the filesystem of the Solr host (a different server) would be complicated, so I have avoided doing so for now.  There will be some deployments that require coordination between the maintainers of the automation and Krikri files, the way that there are now with other parts of our system.  Furthermore, the schema file is really part of Solr, so, I think, should stay with that role.  If Solr wants to change the schema options with a new release, I'd rather have the file, and its change history, part of the automation project.
- We are now running different versions of Ubuntu:  12.04 LTS on the original servers and 14.04 LTS on the new ingestion2 servers.  (See the Vagrantfiles.)  So far, so good; installation has not been a problem, except for easily-remedied issues with a Redis confguration file.  I don't believe that having separate versions of the same OS should be any problem, but you might have observations to the contrary, or good questions that I haven't thought of, so let me know.  We've only been running 12.04 on the other servers because of a BigCouch dependency, but that will be going away soon.

TODO
--------
~~Before this PR is merged I will add dedicated deployment playbooks.  Deployments are currently done by inclusion of the deployment task file by the main role's task file, which means that you have to run everything for the ingestion_app role.~~  I will add dedicated deployment playbooks in the near future.  The existing `ingestion_app_*` playbooks are good enough for now.

I have not set up Munin to monitor these new servers, but munin-node is installed and seems to be running, as part of the "common" role.


ansible/roles/dev_loadbalancer/templates/haproxy.cfg.j2
```diff
@@ -16,6 +16,7 @@ defaults
 	timeout client 50s
 	timeout server 5m
 
+{% if ingestion2 is not defined %}
```
anarchivist: I'm trying to understand what happens in this `if...endif` block. Do the load balancers for BigCouch and Elasticsearch don't get defined if `ingestion2` is not defined?

anarchivist: (I realize it may make more sense once I review this more.)

markbreedlove: Yes, because with ingestion 2 we're no longer using BigCouch.  Also, this being the _dev_ loadbalancer for ingestion 2, we're only running one Elasticsearch node, so there's no need for the loadbalancer.  We're running one node to save memory, because the focus is on testing ingestion2.

ansible/roles/ingestion_app/templates/config_initializers_redis.rb.j2
```diff
@@ -0,0 +1,5 @@
+
+{# Assume one Redis server for now #}
+{% set redis_host = groups['redis'][0] %}
+
+Resque.redis = 'redis://{{ redis_host }}:6379'
```
anarchivist: This might be nitpicky, but should we just change Heidrun to use RailsConfig to set a variable for the environments in, say, `config/redis.yml` (see [here](http://dchua.com/2013/07/30/howto-multiple-redis-database-for-environment/), for example).

ansible/roles/dev_loadbalancer/templates/haproxy.cfg.j2
anarchivist: Sidebar: until we implement migrate the API key database and make the changes necessary to point to wherever the keys will be stored, we'll still need BigCouch.

README-ingestion2.md
```diff
@@ -0,0 +1,61 @@
...
+```
+
+The various sites will be online at:
+
+* http://local.dp.la/
+* http://local.dp.la:8080/v2/items (the API)
+* http://ingestion2.local.dp.la/  (Heiðrún)
+* http://ingestion2.local.dp.la:8080/manager/html (Tomcat admin interface)
+* http://ingestion2.local.dp.la:8080/solr/#/ (Solr admin interface)
+* http://ingestion2.local.dp.la:8080/marmotta/ (Marmotta admin interface)
```
no-reply: We will need to put marmotta (particularly its LDP endpoint) online at an address that we're okay with being the base namespace for our LDP items.

ansible/roles/ingestion_app/templates/etc_init.d_unicorn_heidrun.j2
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
+### BEGIN INIT INFO
+# Provides: unicorn_heidrun
+# Required-Start: $all
+# Required-Stop: $network $local_fs $syslog
+# Default-Start: 2 3 4 5
+# Default-Stop: 0 1 6
+# Short-Description: Start/stop Heidrun unicorn
```
anarchivist: At some point we'll probably want to review all of Unicorn init scripts and see if we can refactor them into a more generic script that takes arguments if it can. Alternately, we could see if we could use a more generic template for an init script that could get extended as needed. (I don't think that needs to happen before we merge this, though.)

ansible/roles/ingestion_app/templates/config_initializers_redis.rb.j2
markbreedlove: I'd usually say yes, and maybe it makes sense to add this environment configuration to the repository, for anyone else trying to use it, but it would add more complexity where changing the setting from its default of "localhost" would usually not be necessary for development. There would be one more file to add here.

ansible/roles/ingestion_app/templates/etc_init.d_unicorn_heidrun.j2
markbreedlove: Yes, I agree.  It's something that's been bugging me.

ansible/roles/ingestion_app/templates/config_initializers_redis.rb.j2
anarchivist: My guess is that we'd move the initializer to Heidrun itself, and then only keep the config files here.

README-ingestion2.md
markbreedlove: Oh, so instead of `/marmotta`?  I'll look at KriKri and see what it's configuring in the engine cart generator.

ansible/roles/dev_loadbalancer/templates/haproxy.cfg.j2
markbreedlove: Ouch, I forgot about that.  Thinking ...

ansible/roles/marmotta/templates/system-config.properties.j2
```diff
@@ -0,0 +1,157 @@
...
+security.permission.default_local.methods = GET
+security.permission.default_local.methods = HEAD
+security.permission.default_local.methods = OPTIONS
+security.permission.default_local.methods = POST
+security.permission.default_local.methods = PUT
+security.permission.default_local.methods = DELETE
+security.permission.default_local.host = LOCAL
+security.permission.default_local.priority = 5
+security.configured = true
+kiwi.context = http://{{ inventory_hostname }}:8080/marmotta/
```
anarchivist: Re: @no-reply's [comment](https://github.com/dpla/automation/pull/51#discussion_r23460059):

> We will need to put marmotta (particularly its LDP endpoint) online at an address that we're okay with being the base namespace for our LDP items.

See the first paragraph or so here: http://marmotta.apache.org/configuration.html

> Note that the host name and port you are using for the first access to your installation is important since decides on how Linked Data resources will be handled later. If you plan a production deployment, you should therefore directly deploy on the server you are going to use and access it via the host name it will have in the future, or manually set the following settings:
>
```
kiwi.context = http://marmotta.example.org/
kiwi.host = http://marmotta.example.org/
kiwi.setup.host = true
```
> where `kiwi.context` indicates the base URI that Marmotta will consider for local resources, `kiwi.host` is the base URL to access the admin interface, and `kiwi.setup.host = true` avoids getting the host configuration overwritten. In case you are running Marmotta behind a proxy, `kiwi.context` should contain the public URI, and you can keep the internal one used as `kiwi.host`.

README-ingestion2.md
anarchivist: See [my comment on the Marmotta `system-config.properties` template](https://github.com/dpla/automation/pull/51/files#r23461114).

no-reply: Our jettywrapper stuff leaves it at localhost/marmotta.  I'm just calling out that we need to get that configuration in Marmotta, and have the URIs actually resolve before we run any *real* harvests.

ansible/roles/dev_loadbalancer/templates/haproxy.cfg.j2
markbreedlove: The API app has some `api_auth` settings in `v1/config/dpla.yml` that can be used in development to circumvent API key checking, so I wonder if I could just use that for the dev environment to avoid having to install BigCouch.

We'll also have to fix the API now rather that later to fetch individual items (by ID) from Elasticsearch instead of BigCouch, but I think we decided that that was going to be pretty easy.

ansible/roles/ingestion_app/files/copy_local_app.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+# Run as root.
+# Copies _both_ heidrun and krikri directories owned by `vagrant' into
+# directories owned by `dpla'
+
+if [ ! -d /krikri ]; then
```
no-reply: Can you say something about what's going on here? I'm thinking the production system shouldn't need to know about Krikri, since that dependency is managed in `heidrun`'s Gemfile.

markbreedlove: Yes, `copy_local_app.sh` is only executed in the context of a "local", development deployment.  In production, the flag that governs "local" installation would be false and this would not be run.

ansible/roles/marmotta/templates/system-config.properties.j2
```diff
@@ -0,0 +1,157 @@
+database.h2.url = jdbc:h2:/v1/marmottadata/marmotta;MVCC=true;DB_CLOSE_ON_EXIT=FALSE;DB_CLOSE_DELAY=10
```
anarchivist: Marmotta database configuration for production should be deployed using Postgres. See [Marmotta's configuration documentation](http://marmotta.apache.org/configuration.html#db) for more details. 

However, we *could* use the H2 embedded database for the development environment, if there are any resource concerns. My suggestion would be that we just stay simple and use Postgres throughout.

ansible/roles/marmotta/tasks/main.yml
```diff
@@ -0,0 +1,23 @@
+---
+
+- file: path=/v1 state=directory owner=root group=root mode=0755
+
+- name: Ensure state of Marmotta database directory
```
anarchivist: See comment below on the Marmotta `system-config.properties` template.

ansible/roles/ingestion_app/templates/config_initializers_redis.rb.j2
no-reply: I'm with @anarchivist on this, and generally against generating production Ruby code at time of deployment.  This initializer is effectively exempt from CI, here, yes?

ansible/roles/solr/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+
+solr_version: 4.10.3
```
anarchivist: Note to self: if this is the case, we should update `marmotta-jetty` with this version of Solr.

ansible/roles/solr/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
...
+
+- name: Ensure the state of the Solr collection config files
+  template: >-
+    src="{{ item }}.j2" dest="/opt/solr/dpla/solr/collection1/conf/{{ item }}"
+    mode=0644 owner=root group=root
+  with_items:
+    - solrconfig.xml
+  notify: restart tomcat
+
+- name: Ensure that Solr schema-related files are up-to-date
```
anarchivist: Would it make more sense to copy these over from the checked out Heidrun repository if we can?

ansible/roles/solr/files/install_solr.sh
```diff
@@ -0,0 +1,81 @@
...
+}
+
+err_exit() {
+    msg=$1
+    >&2 echo $msg
+    exit 1
+}
+
+solr_home_absent() {
+    test -d /opt/solr-$VERSION && return 1
```
anarchivist: Here and in `ensure_solr_home_exists()`, you're testing `/opt/solr-$VERSION`, but it looks like [below in `ansible/roles/solr/tasks/main.yml`](https://github.com/dpla/automation/pull/51/files#diff-7ea063901014a9616e447468c186ad5cR16) you're referring to `/opt/solr`.

ansible/roles/solr/files/schema.xml
```diff
@@ -0,0 +1,755 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
anarchivist: See [comment here](https://github.com/dpla/automation/pull/51/files#r23463012).

ansible/roles/solr/files/stopwords_en.txt
```diff
@@ -0,0 +1,58 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
anarchivist: See [comment here](https://github.com/dpla/automation/pull/51/files#r23463012).

ansible/roles/solr/files/install_solr.sh
```diff
@@ -0,0 +1,81 @@
...
+    for jar in $jars; do
+        ln -sf $solr_home/dpla/lib/ext/$jar /usr/share/tomcat7/lib/$jar \
+            || err_exit "Could not symlink $jar"
+    done
+}
+
+test -z "$VERSION" && usage 1
+
+if solr_user_absent; then
+    useradd -c 'Solr' -d /opt/solr -s /bin/false solr \
```
anarchivist: See [comment here](https://github.com/dpla/automation/pull/51/files#r23463399).

ansible/roles/solr/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
+---
```
anarchivist: See [comment here](https://github.com/dpla/automation/pull/51/files#r23463399).

ansible/roles/solr/templates/solrconfig.xml.j2
```diff
@@ -0,0 +1,1905 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
anarchivist: I'm not sure why this is a template, and why it can't be coped over from Heidrun.

ansible/roles/solr/templates/solrconfig.xml.j2
```diff
@@ -0,0 +1,1905 @@
...
+       have your own custom plugins.
+    -->
+
+  <!-- Controls what version of Lucene various components of Solr
+       adhere to.  Generally, you want to use the latest version to
+       get all bug fixes and improvements. It is highly recommended
+       that you fully re-index after changing this setting as it can
+       affect both how text is indexed and queried.
+  -->
+  <luceneMatchVersion>4.10.3</luceneMatchVersion>
```
anarchivist: Ah, this might be why.

ansible/roles/solr/tasks/main.yml
no-reply: I think we really need to figure this out. As it stands, the solr schema and config are checked into version control in three separate places (`krikri`, `heidrun` (after being added to the repository by the krikri install generator), and here). That seems terribly fragile, and makes it hard to know that the system we're testing is the system we're deploying.

ansible/roles/solr/templates/solrconfig.xml.j2
```diff
@@ -0,0 +1,1905 @@
...
+  -->
+  
+  <!-- Data Directory
+
+       Used to specify an alternate directory to hold all index data
+       other than the default ./data under the Solr home.  If
+       replication is in use, this should match the replication
+       configuration.
+    -->
+  <dataDir>/v1/solrdata</dataDir>
```
anarchivist: Does this need to be set here?

anarchivist: IOW, we can probably set a Solr data directory elsewhere.

ansible/roles/solr/tasks/main.yml
markbreedlove: This was the issue I was referencing in my "controversial" topics section in this PR. :-)

Now that I look at it, it's slightly more complicated, as well, by the fact that both Krikri and Heidrun have their own `solr_conf` directory ... is that just an oversight?

Anyway, grabbing the file from Heidrun or Krikri means a more complex deployment.  I originally thought I'd have to get the schema file from Krikri by examining which version was required by Heidrun ... i'm a little less certain now that I see that Heidrun has its own copies of the files.  Next, I'd have to retrieve just those files from a checkout of the Krikri / Heidrun repo, and then delegate a task to the Solr server to modify the files in Solr's filesystem.  It's just a bit tricky.

Another thing is that these files really are part of Solr, and if something changes from one Solr version to another (I'm not sure how often that happens) I'd want to have these files versioned alongside the rest of the Solr configuration.

I think the main issue was the one above, with the delegation and having the tasks for the Heidrun role reach over and modify the Solr server's files.  That was actually my original plan, but then I backed out because I sensed it was going to get complicated.

no-reply: > Now that I look at it, it's slightly more complicated, as well, by the fact that both Krikri and Heidrun have their own solr_conf directory ... is that just an oversight?

The files exist in each place because they need to be available locally for the `jettywrapper` tasks.  

There is probably a way we could make them accessible to Heidrun without checking them into git, but it would need to be overridable, since the assumption that all `Krikri` users will want the exact schema we are using is not one we would want to make.

> ...and then delegate a task to the Solr server to modify the files in Solr's filesystem.

Is this true of `schema` or just `solrconfig`?  If the former, than I think it's especially important that we sort this out now.  `schema.xml` is definitely *not* a part of solr (at least not the parts that define the actual fields), and really needs to be deployed as it was tested.

I see why this is complicated, but I'm hopeful we can at least figure out a plan so this doesn't become the status quo.

anarchivist: @no-reply said:

> There is probably a way we could make them accessible to Heidrun without checking them into git, but it would need to be overridable, since the assumption that all Krikri users will want the exact schema we are using is not one we would want to make.

Yes, and this is the argument to copy them from Heidrun (somehow). We don't even necessarily need to check out the repo; we can grab the file using the Github raw URL for whatever branch we want.

@markbreedlove said:

> Another thing is that these files really are part of Solr, and if something changes from one Solr version to another (I'm not sure how often that happens) I'd want to have these files versioned alongside the rest of the Solr configuration.

`marmotta-jetty` currently uses Solr 4.9.0. I can make a build that packages up Solr 4.10.3 to test. We should be consistent about which versions we're deploying.

ansible/roles/solr/files/install_solr.sh
markbreedlove: Yes, that's by design.  At this stage of the installation, we need to check that the particualr version of Solr is installed.  `/opt/solr` is a symlink to whatever is the current version's directory of Solr.  The task that follows to ensure the state of the Solr context file will always reference the correct file for the current version, because the symlink is set [earlier](https://github.com/dpla/automation/pull/51/files#diff-63fa04f7324e4e9c53277ca85248e8bdR74).  I chose this because it felt cleaner and more abstract than parameterizing the directory name in that task, and because if `/opt/solr` does not exist, it will error out and flag the issue.

Similarly, I set the `solr` user to `/opt/solr` so it's always current.  Now that I think about it, I think I never really used the `solr` account for anything, so I should remove it.

anarchivist: Got it. Thanks!

anarchivist:
We might want to look at the [Marmotta production deployment info](http://marmotta.apache.org/configuration.html#production) and decide what we want to apply here (either to `development` or regardless of level):

* Firewall configuration
* Connector or reverse proxy (they suggest AJP, but we could perhaps use Nginx)
* Security configuration for Marmotta admin

ansible/roles/ingestion_app/files/copy_local_app.sh
no-reply: Okay, I follow.  I don't really have a strong opinion about this, but I think there may be an argument for deferring to the `Heidrun` developer's Gemfile in a development environment.  

For my part, if I want to be testing heidrun against anything other than a released version of `krikri` I would make that change explicitly.

ansible/roles/ingestion_app/templates/config_initializers_redis.rb.j2
markbreedlove: OK, that's fine.  I'll make two coordinated changes in `heidrun` and `automation`:  In `heidrun`, to add the initializer the reads the YAML file; in `automation`, to update a `settings.local.yml`, I presume.

no-reply:
This looks good to me, overall.  There are a few points where I'd like to iron out the relationship between the `heidrun` repository, `jettywrapper`, `travis-ci` and the deployed code/configuration.

Other than that, I think it's just a matter of getting our Marmotta configuration ironed out.

Thanks for your work on this @markbreedlove. :guitar:

anarchivist:
Agreed with @no-reply. This is great! Let us know if you have any questions about the Marmotta config matters.

ansible/roles/ingestion_app/templates/config_initializers_redis.rb.j2
no-reply: This initializer should probably be in `krikri`, since the dependency on Redis is there.

markbreedlove:
Delegating the fetch of the solr files to a task on the solr server that gets them from Github is actually not so hard, it now occurs to me, so I'll make that happen soon.

markbreedlove:
I'm going to squash this now.  There are some minor things, like additional Marmotta tuning, that can go into new branches in the near future.  I'd like to merge this soon and release it to have it sync with the recent release of Krikri.

PR #52, opened 2015-01-16
Fix BigCouch installation

markbreedlove:
Provide .deb package files to download, because the packages that we
used to fetch from the Cloudant repository have been removed.

Though we're going to stop using BigCouch soon, we have to fix this so that new installations of dev VMs with the legacy system don't fail.

### DIFF
--- a/ansible/roles/dbnode/tasks/main.yml
+++ b/ansible/roles/dbnode/tasks/main.yml
@@ -5,22 +5,54 @@
   tags:
   - database
 
-- name: Add cloudant package list
-  template: src=cloudant.list.j2 dest=/etc/apt/sources.list.d/cloudant.list
+- name: Ensure that runit is installed
+  apt: name=runit state=present
   tags:
     - packages
     - database
 
-- name: Update apt package index for bigcouch
-  apt: update_cache=yes
+- name: Figure out if the Spidermonkey package has to be downloaded
+  stat: path=/var/tmp/libmozjs185-cloudant_1.8.5-cloudant1_amd64.deb
+  register: libmozjs_deb
   tags:
     - packages
     - database
 
-- name: Install packages
-  apt: pkg={{ item }} state=present force=yes
-  with_items:
-    - bigcouch
+- name: Figure out if the BigCouch package has to be downloaded
+  stat: path=/var/tmp/bigcouch_0.4.2-1_amd64.deb
+  register: bigcouch_deb
+  tags:
+    - packages
+    - database
+
+- name: "Download Spidermonkey package, if necessary"
+  get_url: >-
+    url="https://s3.amazonaws.com/dpla-automation-download/libmozjs185-cloudant_1.8.5-cloudant1_amd64.deb"
+    dest="/var/tmp/libmozjs185-cloudant_1.8.5-cloudant1_amd64.deb"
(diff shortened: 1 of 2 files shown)


ansible/roles/dbnode/tasks/main.yml
```diff
@@ -5,22 +5,52 @@
...
+- name: Figure out if the BigCouch package has to be downloaded
+  stat: path=/var/tmp/bigcouch_0.4.2-1_amd64.deb
+  register: bigcouch_deb
+  tags:
+    - packages
+    - database
+
+- name: "Download Spidermonkey package, if necessary"
+  get_url: >-
+    url="https://s3.amazonaws.com/dpla-automation-download/libmozjs185-cloudant_1.8.5-cloudant1_amd64.deb"
```
anarchivist: Can we add a checksum verification here? The `get_url` Ansible module has a `sha256sum` option, which allows us us to pass a SHA-256 for the file.

ansible/roles/dbnode/tasks/main.yml
```diff
@@ -5,22 +5,52 @@
...
+    dest="/var/tmp/libmozjs185-cloudant_1.8.5-cloudant1_amd64.deb"
+  when: not libmozjs_deb.stat.exists
+  tags:
+    - packages
+    - database
+
+- name: "Download bigcouch package, if necessary"
+  get_url: >-
+    url="https://s3.amazonaws.com/dpla-automation-download/bigcouch_0.4.2-1_amd64.deb"
```
anarchivist: Same comment as above.

ansible/roles/dbnode/tasks/main.yml
```diff
@@ -5,22 +5,52 @@
...
+- name: Ensure that runit is installed
+  apt: name=runit state=present
   tags:
     - packages
     - database
 
-- name: Update apt package index for bigcouch
-  apt: update_cache=yes
+- name: Figure out if the Spidermonkey package has to be downloaded
+  stat: path=/var/tmp/libmozjs185-cloudant_1.8.5-cloudant1_amd64.deb
```
anarchivist: I don't know if we want to add another checksum verification based on the `stat` call, but if we do, then it's worth noting that Ansible's options for `stat` include two confusing options:

* `get_md5` (self explanatory; returns boolean, sets `md5` on the returned `stat` object)
* `get_checksum` (calculates a SHA-1; returns boolean, sets `checksum` on the returned `stat` object)

anarchivist:
Other than my comments about checksum verification (which I think should be minor changes) I think this is good to go. :+1:

ansible/roles/dbnode/tasks/main.yml
markbreedlove: I've added checksum checking on line 32 and below on line 42.

anarchivist:
With the checksum verification I think this is ready to :squash:.

ansible/roles/dbnode/tasks/main.yml
markbreedlove: I think we're OK checking it when it's downloaded.  It should be safe to assume that if it exists here in /var/tmp, it's because it was downloaded by the `get_url` module, which checked the checksum.

anarchivist:
:rocket:

PR #53, opened 2015-01-22
Add Travis-CI configuration

anarchivist:
* Checks syntax of playbooks
* Runs ansible-lint on playbooks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,18 @@
+---
+language: python
+python: "2.7"
+sudo: false
+
+install:
+  # Install Ansible.
+  - pip install ansible ansible-lint
+  - pip install -r requirements.txt
+
+script:
+  - echo localhost > ansible/test_inventory
+  - cd ansible
+  - echo Checking syntax of all playbooks ...
+  - ansible-playbook -i test_inventory --syntax-check *_all.yml
+  - echo Running ansible-lint on all playbooks. Rule reference follows.
+  - ansible-lint -L
+  - ansible-lint -q *_all.yml | sort | uniq
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 
 # DPLA Automation
 
+[![Build Status](https://travis-ci.org/dpla/automation.svg?branch=master)](https://travis-ci.org/dpla/automation)
+
 The intention of this project is to provide automated configuration management
 for production, development, and staging environments with the same set of
 files.


markbreedlove:
That's great!  I'm curious to try the lint command independently, but this should be good to merge.

anarchivist:
`ansible-lint` mostly raises warnings AFAIK - I don't think it raises an error otherwise.

no-reply:
:+1: Any objections to my :merge: ?

markbreedlove:
No objections.  I just wanted to try out the ansible-lint command myself before merging, to verify how it behaves when there's a serious parsing issue.

no-reply:
Ah, I see.  I'll let you go ahead and do that.

.travis.yml
```diff
@@ -0,0 +1,18 @@
...
+  - pip install -r requirements.txt
+
+script:
+  - echo localhost > ansible/test_inventory
+  - cd ansible
+  - echo Checking syntax of all playbooks ...
+  - ansible-playbook -i test_inventory --syntax-check *_all.yml
+  - echo Running ansible-lint on all playbooks. Rule reference follows.
+  - ansible-lint -q *_all.yml | sort | uniq
+  - ansible-lint -L
```
markbreedlove: I notice that ansible-lint returns an error status (2) when there are issues.  If these two lines were reversed, I'm assuming that the CI run would return an error status as a result of the last statement returning non-zero.  Would that be more desirable?  (The last line, with -L always succeeds.)

anarchivist: I'm fine reordering the lines. The linter's rules are as follows:

```
ANSIBLE0001: Old style (${var}) brackets
  Checks for old style ${var} rather than {{var}}
ANSIBLE0002: Trailing whitespace
  There should not be any trailing whitespace
ANSIBLE0003: Mismatched { and }
  If lines contain more { than } or vice versa then templating can fail nastily
ANSIBLE0004: Git checkouts must contain explicit version
  All version control checkouts must point to an explicit commit or tag, not just "latest"
ANSIBLE0005: Mercurial checkouts must contain explicit revision
  All version control checkouts must point to an explicit commit or tag, not just "latest"
ANSIBLE0006: Using command rather than module
  Executing a command when there is an Ansible module is generally a bad idea
ANSIBLE0007: Using command rather than an argument to e.g. file
  Executing a command when there is are arguments to modules is generally a bad idea
```

In my opinion the only issue I can really see as potentially braking things is `ANSIBLE0003`; the other things are more stylistic concerns. Let me know what you think.

markbreedlove: Yes, as it stands, it would fail due to a preexisting file with extra whitespace, a relatively minor fault.  I'm happy to fix that, however, and have it fail everything on the list above, so we're aware of it.  It seems to be pretty smart about the "Using command ..." rules, and doesn't fail those unnecessarily, such that the builds should pass if I fix that whitespace.

markbreedlove:
This looks great.  If there's an actual parse error, `ansible-playbook --syntax-check` will obviously catch it, but `ansible-lint` will do the same.  It's good as it is, though, with the `--syntax-check` run distinguishing a parse error from other issues.  I will rebase this and merge it.

markbreedlove:
Actually, one final thing.  I think we want to add the travis build-status badge to the README.  I'll look into that, as well.

PR #54, opened 2015-01-23
Ensure that Cloudant package list has been removed

markbreedlove:
If the old apt source list file is still present, it will cause
errors with package upgrades.  The file used to be necessary for our
BigCouch installation and might be present on systems that have existed
before we removed it.


anarchivist:
:+1:

PR #55, opened 2015-02-04
Make PostgreSQL maintenance playbooks more flexible

markbreedlove:
Allow for manual initiation and disabling of maintenance mode outside of a reboot.

Also add clearing of proxy cache, which was necessary but missing.

This is a minor change I had hanging around and had forgotten about until now, but will be convenient to have merged before making some upcoming changes to production.  It makes it more convenient to put the frontend into maintenance mode for something other than a server reboot.  It also fixes the fact that the proxy cache was not being cleared.


anarchivist:
:100:

PR #57, opened 2015-02-17
Add standalone ingestion app deployment playbooks

markbreedlove:
This adds discrete playbooks for deploying the ingestion application, without performing the role's other provisioning tasks.


anarchivist:
:+1:

PR #59, opened 2015-02-22
Support new vendored deployment for mappings

anarchivist:
See dpla/heidrun#21. I still need to test this, but I wanted to push up and share my changes. I'll try to test this later today.

### DIFF
--- a/ansible/roles/ingestion_app/files/copy_local_app.sh
+++ b/ansible/roles/ingestion_app/files/copy_local_app.sh
@@ -14,11 +14,20 @@ if [ ! -d /heidrun ]; then
     exit 1
 fi
 
+if [ ! -d /heidrun-mappings ]; then
+    >&2 echo "Local directory /heidrun-mappings does not exist"
+    exit 1
+fi
+
 if [ ! -d /home/dpla/krikri ]; then
     mkdir /home/dpla/krikri
 fi
 if [ ! -d /home/dpla/heidrun ]; then
     mkdir /home/dpla/heidrun
+fi
+
+if [ ! -d /home/dpla/heidrun-mappings ]; then
+    mkdir /home/dpla/heidrun-mappings
 fi
 
 rsync -rptl --delete --checksum \
@@ -30,4 +39,7 @@ rsync -rptl --delete --checksum \
     /heidrun/ /home/dpla/heidrun \
     || exit 1
 
-
+rsync -rptl --delete --checksum \
+    --exclude '.git*' --exclude 'README.md' --exclude 'LICENSE' \
+    /heidrun-mappings/ /home/dpla/heidrun-mappings \
+    || exit 1
--- a/ansible/roles/ingestion_app/tasks/deploy.yml
+++ b/ansible/roles/ingestion_app/tasks/deploy.yml
@@ -84,6 +84,16 @@
   tags:
     - deployment
 
+- name: Check out mappings from their repository (not using local source)
+  git: >-
+      repo=https://github.com/dpla/heidrun-mappings.git
+      dest=/home/dpla/heidrun-mappings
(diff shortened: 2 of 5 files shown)


ansible/roles/ingestion_app/files/copy_local_app.sh
```diff
@@ -30,4 +39,9 @@ rsync -rptl --delete --checksum \
     /heidrun/ /home/dpla/heidrun \
     || exit 1
 
-
+# Don't delete "extraneous" files in the destination directory, unlike the
+# other two rsync calls
+rsync -rptl --checksum \
+    --exclude '.git*' --exclude 'README.md' --exclude 'LICENSE' \
+    /heidrun-mappings /home/dpla/heidrun/vendor/mappings \
```
markbreedlove: This will copy it into `/home/dpla/heidrun/vendor/mappings/heidrun-mappings` because there's no ending '/' after `/heidrun-mappings`.  Is the intention to copy it into this nested structure?  Also, the directory that's checked above is `/home/dpla/heidrun-mappings`, as opposed to the vendor directory.

ansible/roles/ingestion_app/tasks/deploy.yml
```diff
@@ -84,6 +84,17 @@
   tags:
     - deployment
 
+- name: Check out mappings from their repository (not using local source)
+  git: >-
+      repo=https://github.com/dpla/heidrun-mappings.git
+      dest=/home/dpla/heidrun-mappngs
```
markbreedlove: Typo:  `mappngs`

ansible/roles/ingestion_app/files/copy_local_app.sh
```diff
@@ -30,4 +39,9 @@ rsync -rptl --delete --checksum \
     /heidrun/ /home/dpla/heidrun \
     || exit 1
 
-
+# Don't delete "extraneous" files in the destination directory, unlike the
```
markbreedlove: Why shouldn't `--delete` be used?  If someone changes a filename, the old one will remain.

anarchivist: This was an oversight, because I wanted to retain the `:heidrun_sample` mapping. I'll fix this up in another commit.

ansible/roles/ingestion_app/tasks/deploy.yml
anarchivist: Fixing, thanks. Just ran into a problem with this too.

ansible/roles/ingestion_app/files/copy_local_app.sh
anarchivist: This will be fixed shortly.

anarchivist:
On my count, this is ready to go. @markbreedlove - feel free to take another gander; @no-reply, feel free to take a look as well.

markbreedlove:
I will test it locally ...

markbreedlove:
I may have to fix this elsewhere, but I get this error if I've already used a local source directory and then switch to the GitHub repository:
```
TASK: [Check out mappings from their repository (not using local source)] ***** 
failed: [dev2] => {"cmd": "/usr/bin/git clone --origin origin --recursive --branch master https://github.com/dpla/heidrun-mappings.git /home/dpla/heidrun-mappings", "failed": true, "rc": 128}
stderr: fatal: destination path '/home/dpla/heidrun-mappings' already exists and is not an empty directory.

msg: fatal: destination path '/home/dpla/heidrun-mappings' already exists and is not an empty directory.
```

I'm going to test something else and get back to you.

markbreedlove:
OK, this appears to work fine.

The issue above with the directory existing already is something I need to fix in its own PR, and I'll do that soon.  I have some changes I'll stash and commit in their own branch.

PR #60, opened 2015-02-23
Improve cooperation between local and git deployments

markbreedlove:
Fix the population of build directories during deployments in the
development environment to avoid errors in some cases when switching
between local-filesystem and git deployments.

This also preserves separate directories for git and local build
directories so that they can be better analyzed independently of
one-another and so that they should be faster to rsync upon subsequent
deployments.


anarchivist:
:cake:

PR #62, opened 2015-03-05
Add param for controlling Munin alert sending

markbreedlove:
Add munin_always_send and munin_always_send_priorities, which govern
whether Munin sends repeated alert messages if the alert's state has
not changed.


anarchivist:
:bell: :no_bell: :ok_hand:

PR #64, opened 2015-03-25
Set Linux kernel params for memory overcommitment

markbreedlove:
See http://www.win.tue.nl/~aeb/linux/lk/lk-9.html
See http://www.etalabs.net/overcommit.html

The default for vm.overcommit_memory is the Linux kernel default of 0,
for the standard heuristic.  The ingestion servers use 2, to allow them
to go a little crazier with malloc and, I suppose, potentially forking
new processes.

The development VM `dev2` is given an overcommit_ratio of 95 to allow it to allocate a little more memory, because it doesn't have a swap partition.  (Some memory is reserved for kernel buffer and block caches.)  I'm not sure if this is the best setting, but let's see how it works.

### DIFF
--- a/ansible/group_vars/ingestion_app
+++ b/ansible/group_vars/ingestion_app
@@ -1,0 +1,2 @@
+# See roles/common/defaults/main.yml
+kernel_vm_overcommit_memory: 2
--- a/ansible/group_vars/ingestion
+++ b/ansible/group_vars/ingestion
@@ -1,0 +1,2 @@
+# See roles/common/defaults/main.yml
+kernel_vm_overcommit_memory: 2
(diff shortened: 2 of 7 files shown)


ansible/group_vars/ingestion_app
```diff
@@ -0,0 +1,2 @@
+# See roles/common/defaults/main.yml
+kernel_vm_overcommit_memory: 2
```
anarchivist: @markbreedlove Are we running into memory issues in the new ingestion app as well?

markbreedlove: No, but it's worthy of this setting because it's the most likely system to have memory issues.  This change anticipates issues like those we've had, as we add more enrichments and enrichment-related functionality, some of which could use libraries that we're unfamiliar with, which may have bad memory-usage patterns.

I've tried to leave systems alone with the default "0" setting wherever possible, just on principle.  Other systems that use a lot of memory have other means of controlling memory usage, like Java's '-X' parameters, and PostgreSQL's many memory-related settings.  Those systems are also more predictable and relatively well-behaved.

anarchivist: OK, duly noted!

PR #66, opened 2015-04-05
Heidrun: add base URI for providers to settings

anarchivist:


markbreedlove:
Looks pretty simple.  Can you verify whether it's safe for this to be hardcoded?  There is a `frontend_hostname` variable that we may be able to use here.

anarchivist:
@markbreedlove The URIs for the providers are not yet dereferenceable, and should be the same across both dev, staging, and production. I think it should be safe for this value to be hardcoded.

markbreedlove:
@anarchivist: OK!  In that case, I will merge this.

anarchivist:
:clap:

PR #67, opened 2015-04-06
Marmotta: disable LDCache module

anarchivist:
Disables the LDCache module, which attempts to fetch remote resources through dereferencing. This will reduce the number of outgoing requests from our Marmotta instance, and may speed up save and query performance.

LDCache is known to at least have some performance implications, as described [here](https://github.com/apache/marmotta/blob/develop/platform/ldcache/marmotta-ldcache-common/src/main/resources/config-defaults.properties#L23-L25).

### DIFF
--- a/ansible/roles/marmotta/templates/system-config.properties.j2
+++ b/ansible/roles/marmotta/templates/system-config.properties.j2
@@ -177,3 +177,4 @@ user.admin.roles = editor
 user.admin.roles = manager
 user.admin.roles = user
 marmotta.home = {{ marmotta_home }}
+ldcache.enabled = false


markbreedlove:
Interesting how they say enabling it should be "considered with care" but it's enabled by default.

Have you measured any difference with this parameter change?

anarchivist:
I have not measured it, but the timeout on these requests was very high (~10 seconds), and it reattempts the requests several times. @no-reply may know more.

anarchivist:
Also, we can talk more about this at today's checkin in case you're curious, but the long-term strategy will probably be to see if there's a way to change the caching behavior to "whitelist" certain URI patterns, or only the objects of of a triple specified by a whitelisted property.

PR #68, opened 2015-04-08
Use file-based caching for API in contentqa

anarchivist:
The API cache in contentqa should be clearable quickly; this moves the caching to use file-based caching.


markbreedlove:
Looks good.  Does the location of the file store need to be documented, and is it something that is configurable?  I am OK merging this, but just wondering.

anarchivist:
At this point, it's hardcoded into the platform app (it always was). It's in `tmp/api-cache` in the application's root.

anarchivist:
:squash:ed and ready to merge, @markbreedlove

PR #70, opened 2015-04-13
Improve Tomcat logfile rotation

markbreedlove:
- Prune access_log logs that were not being deleted
- Add parameters for age of "catalina" and AccessLogValve logs
- Change "catalina" logs to rotate daily, instead of weekly


no-reply:
This looks good to me. :rocket:

anarchivist:
:+1:

PR #73, opened 2015-04-21
[site proxy] Enable CORS for API redirects

anarchivist:
Complements dpla/platform#37. This is necessary, since our JSON-LD `@context` URIs are referred to using `http://dp.la/api`.

### DIFF
--- a/ansible/roles/site_proxy/templates/etc_nginx_sites_available_site-proxy.j2
+++ b/ansible/roles/site_proxy/templates/etc_nginx_sites_available_site-proxy.j2
@@ -152,11 +152,15 @@ server {
         rewrite ^/about/map /info/map last;
         rewrite ^/about/schema /info/schema last;
         rewrite ^/platform /info/developers/codex last;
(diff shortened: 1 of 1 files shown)


markbreedlove:
Ooh, right.  I will test this now, if you haven't.

markbreedlove:
In order to prevent unlikely but possible errors like the following, I think that the rewrite `location` should specify a `/` after `/api`, as in `^/api/(.*)` and then have the `rewrite` line look like `rewrite ^/api/(.*)$ http://{{ api_hostname }}/$1 last;`  Here's' what can happen now:
```
$ curl -i -s local.dp.la/apifoo
HTTP/1.1 302 Moved Temporarily
[...]
Location: http://local.dp.lafoo
```

anarchivist:
Force-pushed with the suggested changes. Thanks, @markbreedlove.

markbreedlove:
OK, that works.  I'll merge it as soon as Travis is done.

PR #83, opened 2015-06-01
Add task to ensure common Munin plugins are enabled

markbreedlove:
Add a task to ensure that specific Munin plugins will always be
installed.

Two network plugins have been explicitly enabled, which are enabled by
default on Ubuntu 14.04, but not 12.04.


anarchivist:
:+1:

PR #84, opened 2015-06-02
Minor Fixes for a From-scratch Run

no-reply:
This fixes a handful of minor issues I ran into when building the automation project from scratch today.

  - Removes a duplicate bullet point from `README.md`
  - Forces Vagrant to use the default `insecure_private_key`. Some versions overwrite that key on new machines by default, breaking Ansible's SSH initial login as the `vagrant` user.
  - Fixes an error with the elasticsearch jobs in `development`.

### DIFF
--- a/Vagrantfile.dist
+++ b/Vagrantfile.dist
@@ -19,8 +19,11 @@ dbnodes = {
 }
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
+    # Use the default insecure_private_key so ansible doesn't need to 
+    # know about all the generated ones.
+    config.ssh.insert_key = false
 
-	# All boxes are hashicorp/precise64, which is Ubuntu 12.04 LTS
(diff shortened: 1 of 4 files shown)


Vagrantfile.dist
```diff
@@ -19,8 +19,11 @@ dbnodes = {
 }
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
+    # Use the default insecure_private_key so ansible doesn't need to 
+    # know about all the generated ones.
+    config.ssh.insert_key = false
```
anarchivist: Does this also need to be in `Vagrantfile.ingestion2`?

markbreedlove:
as @anarchivist mentioned, `Vagrantfile.ingestion2` should also have the `config.ssh.insert_key` change, but, other than that, this is good to merge if that can be added and squashed with commit 88e4906.

PR #85, opened 2015-06-08
Add queues to development VM inventory

markbreedlove:
Add queues for mapping, enrichment, and indexing.
Allow just one worker for each queue.


anarchivist:
:+1:

PR #87, opened 2015-06-08
Add HOME environment var to god config

markbreedlove:
Have god start resque workers with the $HOME environment variable
correctly set.

I noticed that my workers in my development VMs were not able to start
without this being set.  There were errors in
/opt/heidrun/log/worker.log indicating:
  ArgumentError: couldn't find HOME environment -- expanding `~'


anarchivist:
:+1:

PR #88, opened 2015-06-09
Improve Solr index and configuration operations

markbreedlove:
- Delete Solr index when there is a schema change.
- Reload the Solr core instead of restarting Tomcat.

There is no standalone playbook for clearing the Solr index, which would
be nice, but which would take longer to put together than I can afford
right now, because I'd have to spend more effort to avoid duplicating
code in that case.

After clearing the index, you have to run a new indexing operation,
which is out of the scope of the `automation` project.  Just note that
that has to happen.  It seems that a first-time indexing run after the
schema is changed, the index is deleted, and the Solr core is restarted
requires a reload of the core again for the added documents to show up.
I don't have an explanation for that, and I'm not sure whether it has to
do with deleting the index, or something that should be done when the
records are re-indexed.

### DIFF
--- a/README-ingestion2.md
+++ b/README-ingestion2.md
@@ -72,6 +72,22 @@ infrastructure.
 Example:
 `ansible-playbook -u mb -i ingestion playbooks/deploy_ingestion_app_development.yml --extra-vars="fast_deployment=true"`
 
+### Heiðrún deployments with Solr changes
+
+Sometimes a Heiðrún deployment will carry along with it some Solr schema or
+configuration changes.  When this happens, the Solr search index that's used
+for QA will have to be deleted and re-created.  An Ansible task will be
+triggered to delete the index, but you will have to re-index your repository
+records yourself.  This is not automated yet, and will have to be done in the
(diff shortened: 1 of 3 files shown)


markbreedlove:
I'm going to add the comments in the PR description to a file in the project.

anarchivist:
I'm :+1:. Maybe @no-reply could also take a look?

no-reply:
I may be missing something. Is there a way to run the deployment without deleting the index?

anarchivist:
My understanding is that the index will only be deleted if the schema changes (see [here](https://github.com/dpla/automation/pull/88/files#diff-de1a36c5fd3eb2b6221e53e7abd7ddbeL21)).

markbreedlove:
Yes, the index deletion happens in a handler, which means it only happens when the schema file has actually changed.

no-reply:
:rocket: thanks.

markbreedlove:
In general, handlers only happen when the task registered "changed", and one can define what constitutes change.

PR #89, opened 2015-06-10
Enable deployment of exhibitions from local source

AudreyAltman:
This addresses ticket #7822 https://issues.dp.la/issues/7822

### DIFF
--- a/ansible/roles/omeka/files/copy_local_app.sh
+++ b/ansible/roles/omeka/files/copy_local_app.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+# Run as root.
+# Copies directory owned by `vagrant' into directory owned
+# by `dpla'
+
+if [ ! -d /exhibitions_dev ]; then
+  echo "Local directory /exhibitions_dev does not exist" >&2
+  exit 1
+fi
+
+if [ ! -d /home/dpla/exhibitions-local ]; then
+  mkdir /home/dpla/exhibitions-local
+fi
+
+rsync -rptl --delete --checksum \
+    --exclude 'application/logs' \
+    /exhibitions_dev/ /home/dpla/exhibitions-local
+if [ $? -ne 0 ]; then
+  exit 1
+fi
+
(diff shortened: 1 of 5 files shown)


AudreyAltman:
I tested this locally by deploying a local branch of the exhibitions app, and it seems to be working as expected.

anarchivist:
I'm :+1:, but maybe have @markbreedlove take a look too before we merge?

ansible/roles/omeka/files/copy_local_app.sh
```diff
@@ -0,0 +1,23 @@
...
+  echo "Local directory /exhibitions_dev does not exist" >&2
+  exit 1
+fi
+
+if [ ! -d /home/dpla/exhibitions-local ]; then
+  mkdir /home/dpla/exhibitions-local
+fi
+
+rsync -rptl --delete --checksum \
+    --exclude 'application/logs' \
```
markbreedlove: Can you confirm that `application/logs` gets created by Omeka when it's started?  In other roles, like `api`, there's a build script that ensures that such a directory exists in the destination, e.g. `/srv/www/api`, when the "build" script is run to copy the app from the home directory.  If it got created in your dev environment, then this should be OK.

markbreedlove:
@AudreyAltman, just fixed something in https://github.com/dpla/automation/commit/3d9eb56688e1677c8b83996c9e1763a4c0a63a47 related to my last comment, and related to the rsync error that you reported to me before.

anarchivist:
I'm :+1: if @markbreedlove's good on this. Shall we merge?

markbreedlove:
I'm fine with merging it.  @AudreyAltman, would you like to squash it first?

ansible/roles/omeka/files/copy_local_app.sh
AudreyAltman: Yes, I can confirm that `application/logs` was created by Omeka.

AudreyAltman:
:squash:ed and ready to :merge: @markbreedlove or @anarchivist

PR #91, opened 2015-06-24
Set PostgreSQL random_page_cost

markbreedlove:
Use a parameter to set random_page_cost.  This is a minor change that
won't result in a major impact, but it's easy to make.  A lower
random_page_cost is more appropriate for SSD drives or Amazon EBS
volumes.

### DIFF
--- a/ansible/roles/postgresql/defaults/main.yml
+++ b/ansible/roles/postgresql/defaults/main.yml
@@ -5,5 +5,8 @@ pg_max_connections: 50
 pg_shared_buffers: 24MB
 pg_effective_cache_size: 128MB
 pg_work_mem: 1MB
+# The default random_page_cost of 4.0 assumes you're using disks.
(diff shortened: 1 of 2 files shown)


ansible/roles/postgresql/defaults/main.yml
```diff
@@ -5,5 +5,8 @@ pg_max_connections: 50
 pg_shared_buffers: 24MB
 pg_effective_cache_size: 128MB
 pg_work_mem: 1MB
+# The default random_page_cost of 4.0 assumes you're using disks.
+# Decrease it if you're using SSD drives.  1.5 might be a good value.
+pg_random_page_cost: 4.0
```
anarchivist: Is this something we're expecting to set to a lower value in our AWS overrides?

markbreedlove: Yes.  I have that ready to go if this is approved.

PR #92, opened 2015-06-25
Disable Marmotta reasoner

anarchivist:
Per discussion on 2015-06-25, we are disabling the reasoner module for Marmotta,
which will allow us to partition Marmotta's database tables more effectively.

(cc @markbreedlove - please add any salient details)

### DIFF
--- a/ansible/roles/marmotta/templates/system-config.properties.j2
+++ b/ansible/roles/marmotta/templates/system-config.properties.j2
@@ -178,3 +178,4 @@ user.admin.roles = manager
 user.admin.roles = user
 marmotta.home = {{ marmotta_home }}
 ldcache.enabled = false
+reasoning.enabled = false


markbreedlove:
This is good.

I'm going to do a full writeup of this elsewhere, but the problem is basically:

I'd like to try partitioning `triples`.  Though it's not ideal, I may be able to get away with partitioning it based on the modulo of the `id` value, breaking it up into what are effectively hash buckets based on the id number.  (I can discuss what I'd do ideally in another location.)

A problem arises in that you can not have a foreign key from another table onto a column of the master table, because the latter is empty.  Foreign keys only work when they point directly to a relation that has data (e.g. a child table).

There are four tables that have foreign keys with `REFERENCES triples(id)`: reasoner_just_supp_triples (0 rows), reasoner_justifications (0 rows),  versions_added (143M rows), versions_removed (8M rows).  

`versions_added` and `versions_removed` could be partitioned on triples_id, with foreign keys for each child partition pointed at the appropriate partition of triples.  I'm not sure if this will work out well, but I want to try it.

However, though all of our `reasoner_*` tables have 0 rows, partitioning them  will be impossible because of a foreign key constraint between two of _them_.  Suppose you partitioned `reasoner_justifications` on `triples_id`.  `reasoner_just_supp_rules` has a foreign key onto `reasoner_justifications.id`.  If `reasoner_justifications` was partitioned by `triples_id`, you would not be able to know, for any `reasoner_just_supp_rules.justification_id`, which partition of `reasoner_justifications`  to point to.

anarchivist:
@markbreedlove we may also want to have a follow up conversation once @no-reply is back about whether we want to keep versioning enabled, if it ends up being a massive performance concern. i know we wanted to hedge our bets on versioning as long as we can, but it may be a fish or cut bait decision.

markbreedlove:
Ah, right.  Then we wouldn't have any tables at all with foreign keys onto `triples`.  Yes, let's talk about that ...

PR #95, opened 2015-07-01
Change bundle install to bundle update

markbreedlove:
In scripts that update Ruby gem bundles, use `bundle update` instead of
`install`, in order to upgrade versions of gems.

See http://bundler.io/v1.3/man/bundle-update.1.html


no-reply:
:rock:

PR #96, opened 2015-07-07
Set max. request body size, Marmotta NGINX proxy

markbreedlove:
Set the maximum request body size, client_max_body_size, in the virtual host of the NGINX reverse proxy in front of Marmotta.  The NGINX default of 1MB has been preventing larger records from being PUT.

See the comment in the template about other Tomcat parameters, but this is at least a step.  After the change, I am able to harvest a record that used to be refused due to its large size.  I've set the maximum request body size to 8 MB, which I think is pretty generous.  I haven't been able to determine Tomcat's default maximum request body size for a PUT request.

### DIFF
--- a/ansible/roles/marmotta/templates/etc_nginx_sites-available_marmotta.j2
+++ b/ansible/roles/marmotta/templates/etc_nginx_sites-available_marmotta.j2
@@ -9,6 +9,16 @@ server {
     server_name {{ inventory_hostname }};
 {% endif %}
 
+    # Maximum request body size.  (Controls maximum size of
+    # a PUT or POST to Marmotta.)
+    # It's not clear whether this is sufficient, or if there's something
+    # that should also be set in Tomcat, e.g. maxPostSize (which, if you
+    # take it literally, does not apply to PUT) or
+    # //servlet/multipart-config/max-request-size inside of a web.xml file.
+    # The web.xml contained inside of marmotta.war does not have a
+    # //servlet element that would pertain.
+    client_max_body_size 8m;
+
     location / {
         # Proxy everything to Marmotta.
         # E.g. http://ldp.somedomain.org/ --> http://thishost:8080/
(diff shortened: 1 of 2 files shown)


ansible/roles/marmotta/templates/etc_nginx_sites-available_marmotta.j2
```diff
@@ -9,6 +9,16 @@ server {
...
 
+    # Maximum request body size.  (Controls maximum size of
+    # a PUT or POST to Marmotta.)
+    # It's not clear whether this is sufficient, or if there's something
+    # that should also be set in Tomcat, e.g. maxPostSize (which, if you
+    # take it literally, does not apply to PUT) or
+    # //servlet/multipart-config/max-request-size inside of a web.xml file.
+    # The web.xml contained inside of marmotta.war does not have a
+    # //servlet element that would pertain.
+    client_max_body_size 8m;
```
anarchivist: It's my understanding that `maxPostSize` only applies to `POST` as well (from the [Tomcat 7 documentation](https://tomcat.apache.org/tomcat-7.0-doc/config/http.html#Common_Attributes)):

> The maximum size in bytes of the POST which will be handled by the container FORM URL parameter parsing. The limit can be disabled by setting this attribute to a value less than zero. If not specified, this attribute is set to 2097152 (2 megabytes).

anarchivist: Also, would it be possible to move `client_max_body_size`'s value into a variable for Ansible?

markbreedlove: I could parameterize `client_max_body_size` but I thought that it should not vary from system to system.  The maximum size of a record that we want to accept, or that Marmotta will allow, will be the same whether the system is development, staging, or production.

When I was looking into the Marmotta and Tomcat configurations, trying to find something related to this, I thought that I'd create a variable and have both NGINX's `client_max_body_size` and the Marmotta / Tomcat settings derived from it, but now it doesn't look like it would help.  It seems that looking up the source of a variable in this particular case would mean one extra step in verifying the allowed size of the request body.

anarchivist: OK, got it - I'm find with this the way it stands.

PR #99, opened 2015-07-24
Rate-limit the bookshelf section of the frontend application

markbreedlove:
This changeset adds a few preliminary changes for parameterizing existing rate-limiting directives, and improving logging.  It then adds rate-limiting directives to place specific controls on the bookshelf, which makes relatively expensive search queries with lots of facets.

### DIFF
--- a/ansible/roles/common_web/defaults/main.yml
+++ b/ansible/roles/common_web/defaults/main.yml
@@ -1,0 +1,3 @@
+---
+
+nginx_log_level: notice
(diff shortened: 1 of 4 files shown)


markbreedlove:
It might take some experimentation to arrive at concurrent-connections and rate settings that are fair to multiple users behind a NAT router (e.g. in a classroom), but that protect against abuse.

Still, even with a classroom full of people, unless they all load a page at the same instant, or scroll the u.i. furiously at the same time, I think it's unlikely that many of them will encounter 503s, due to the configured "burst" value, which simply enqueues requests to be served after current requests are fulfilled.

anarchivist:
Sounds good to me, and we can always tweak this further should need be.

PR #100, opened 2015-07-31
Disable proxy caching of bookshelf

markbreedlove:
TODO: determine whether and how to cache bookshelf responses.

HTML and JSON responses were being cached under the same
proxy_cache_key, leading to JSON being returned for the main bookshelf
page.  If caching is enabled, it should use the suggested
proxy_cache_key value to ensure that JSON and HTML do not get mixed up.

This is related to the frontend's design decision to have the same
resource return not only different representations given different
Accept: content types, but different objects (data), as well.


anarchivist:
:100:

PR #103, opened 2015-08-11
Set dpla account's RAILS_ENV for ingestion app

markbreedlove:
This is significant in production, or wherever the RAILS_ENV should not be the default, `development`.  If you are using the wrong RAILS_ENV (e.g. `development` on a production system), various settings could be off; for example (in my experience) you might enqueue jobs that get sent to the wrong redis queue, where they will not be picked up by the workers, (e.g. which are listening on the `production` queue).

### DIFF
--- a/ansible/roles/ingestion_app/tasks/main.yml
+++ b/ansible/roles/ingestion_app/tasks/main.yml
@@ -31,6 +31,14 @@
   tags:
     - users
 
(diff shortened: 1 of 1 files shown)


no-reply:
:+1: to this.

anarchivist:
:railway_car: :railway_car: :gem: :gem: :deciduous_tree: :evergreen_tree:

PR #108, opened 2015-08-21
Change rsync paths for heidrun-mappings

markbreedlove:
Change the rsync path arguments in the build script for heidrun, so that
heidrun-mappings gets rsynced into place without its top-level
directory's contents.  Only the `heidrun` directory needs to be
transferred, and the changes suggested by
https://github.com/dpla/heidrun-mappings/pull/28 will result in a `spec`
directory that will have files with `require` statements that will
result in load errors.

### DIFF
--- a/ansible/roles/ingestion_app/files/build_ingestion.sh
+++ b/ansible/roles/ingestion_app/files/build_ingestion.sh
@@ -65,10 +65,7 @@ echo "rsyncing mappings ..." >> $LOGFILE
 # Don't delete "extraneous" files in the destination directory, unlike the
 # other two rsync calls
 /usr/bin/rsync -rIptolg --checksum --delay-updates \
-    --exclude 'README.md' \
-    --exclude 'LICENSE' \
-    --exclude '.git*' \
-    /home/dpla/heidrun-mappings/ /opt/heidrun/vendor/mappings
+    /home/dpla/heidrun-mappings/heidrun /opt/heidrun/vendor/mappings
 if [ $? -ne 0 ]; then
     exit 1
 fi


anarchivist:
Have you confirmed that the mappings when deployed get loaded correctly?

markbreedlove:
Yes, I was actually using this change when I was working on `heidrun-mappings` for Florida.

markbreedlove:
This PR is not really that important anymore, in light of the discussion in https://github.com/dpla/heidrun-mappings/pull/28 about moving the `MappingTools` module.

The way presented here is technically a cleaner way of copying the needed directory, and may be a more appropriate model for someone to emulate who is working on mappings outside of our organization.

markbreedlove:
Hey, I just want to give this a bump and revive it.  It's a minor thing, and, as stated above, not strictly necessary, but I think it's a cleaner way of deploying the mappings, and a better example of how it should be done in an installation where there are multiple `vendor` mappings.

If we agree to merge this, it would be nice if I could rebase it before that happens.

ansible/roles/ingestion_app/files/build_ingestion.sh
```diff
@@ -67,7 +67,7 @@ echo "rsyncing mappings ..." >> $LOGFILE
     --exclude 'README.md' \
     --exclude 'LICENSE' \
     --exclude '.git*' \
```
AudreyAltman: Now that the target directory is `heidrun`, instead of `heidrun-mappings`, can you delete the above `--exclude` statements, since those files are not in the `heidrun` directory anyway?

AudreyAltman:
This looks good to me, aside from my one minor question.  Might be good to have @anarchivist or @no-reply take a peek as well, since they are more familiar with the issue.

ansible/roles/ingestion_app/files/build_ingestion.sh
markbreedlove: Yes!  Thanks for catching that.

no-reply:
I think i have rebased this successfully in #142. I didn't want to force push over this branch, just in case my conflict resolution was somehow lossy. 

We can just merge that one, or I can push it to this branch if someone gives it a :+1:.

markbreedlove:
I just wanted to give this a bump to indicate I haven't forgotten. I'm going to assign it to myself and take care of the rebase in this PR some time in the next couple of work days.

markbreedlove:
I've just rebased this branch successfully, so this should be OK to merge once the tests finish.

PR #110, opened 2015-08-24
Add sparql.timeout vars to Marmotta

markbreedlove:
Add sparql.timeout.type and sparql.timeout.description to system-meta.properties.


anarchivist:
I think we also need to add `sparql.timeout = 360` to `system-config.properties`, based upon what's on `marmotta-prod1`.

markbreedlove:
@anarchivist: OK, just added that.

PR #112, opened 2015-08-26
Feature/add primary source sets

markbreedlove:
This should set up the `primary-source-sets` application in your development VM, using the `development` inventory.

## Testing notes

Be sure to update your `Vagrantfile` with the line from `Vagrantfile.dist` that mounts the `/pss_dev` directory.

You'll want to add `pss_branch_or_tag: develop` to `roles/pss/vars/development.yml`, until the `master` branch has been updated with a recent Gemfile change that adds the `pg` gem.  Or use your local source directory.

The new app should be online after deployment at `local.dp.la/primary-source-sets`, but it should give you an error about routes needing to be added.


markbreedlove:
I forgot that I need to add some playbooks for standalone application configuration and deployment.  Working on that.

anarchivist:
I think I am largely :+1:. The routing issue is because the application doesn't know what to do with paths that are prefixed with `/pss`. We might need to fix this at the level of the application.

AudreyAltman:
Ideally I would like the base URI to be http://dp.la/sets - but I'm not sure how to make that happen.

AudreyAltman:
@anarchivist when you say "routing issue", do you mean that when you deploy locally to development, it doesn't recognize `http://local.dp.la/pss`?  Because I might know how to fix that in the `primary-source-sets` app...

anarchivist:
This has to do with the proxy setup that's included in this PR. The proxy is presenting the primary source sets app at `http://{{ frontend_hostname }}/pss`. When you try to access the app there, you get `ActionController::RoutingError (No route matches [GET] "/pss")`:

screen shot 2015-09-01 at 7 43 59 am

The proxy is passing the `/pss` part of the URI to the backend app directly, and no route matches.

In terms of the URI base where it's served from, I think we can be flexible about it - but if it's served from `http://{{ frontend_hostname }}/sets` then individual set URIs may end up look like `http://dp.la/sets/sets/1`.

anarchivist:
I've gone ahead and expanded this to allow for `pss` to deployed into staging. Additional changes include changing the root URL (at Franky and @AudreyAltman's request), using a different variable to set the version of Ruby used by `pss`, adding some more templates and packages, and allowing the `cms` group to connect to the Postgres server.

I'm going to do another run through locally to ensure that deployment works in development, but I think this is ready to squash and merge.

anarchivist:
OK, this is ready to be merged, and has been squashed into two commits. Also, please note that I had to change the reference to `ruby_rbenv_version` to `pss_rbenv_version` since the variables were not getting resolved correctly.

ansible/roles/pss/templates/unicorn.rb.j2
```diff
@@ -0,0 +1,101 @@
...
+#
+# See http://unicorn.bogomips.org/Unicorn/Configurator.html for complete
+# documentation.
+user "dpla", "dpla"
+
+# Use at least one worker per core if you're on a dedicated server,
+# more will usually help for _short_ waits on databases/caches.
+worker_processes {{ unicorn_worker_processes }}
+
+# Since Unicorn is never exposed to outside clients, it does not need to
```
no-reply: Do we want to clean up this file, so the configuration is more readable?

no-reply:
This looks good to me.

If the answer to my inline question is "no way, pal", then :rocket:

AudreyAltman:
Thank you all so much for your hard work on this PR!

no-reply:
:dart:

PR #113, opened 2015-08-26
Revert "Add NGINX keepalive, marmotta proxy" (refs #7903)

anarchivist:
This reverts commit bcc646c3ab06688f034561997728722bbe500245.

dpla/automation#104 (Add NGINX keepalive, marmotta proxy) introduced an issue wherein upon nginx starting, it spits out an error:

```
2015/08/26 19:12:48 [emerg] 21963#0: invalid parameter "max_conns=200" in /etc/nginx/sites-enabled/marmotta:3
```

`max_conns` is only available with Nginx Plus (their commercial offering) based upon the subtle message in the documentation:
<http://nginx.org/en/docs/http/ngx_http_upstream_module.html#down>


markbreedlove:
Oops, that is no good.  I'll follow up some later day on the correct keepalive implementation for our installation, and merge this now.

PR #115, opened 2015-08-27
Clean up stale Unicorn files

markbreedlove:
Remove pid and socket files during deployments, after the shutdown step, in case these files were left around from an earlier failure.  Their presence will cause the startup to fail after the build step is complete.

### DIFF
--- a/ansible/roles/api/tasks/deploy.yml
+++ b/ansible/roles/api/tasks/deploy.yml
@@ -107,6 +107,19 @@
     - api_deployment
     - web
 
+- name: Assert that Unicorn pid and socket files are absent
+  # These could have been left behind by an earlier failure or termination
+  file: >-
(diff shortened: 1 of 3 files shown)


anarchivist:
:+1: Looks good to me. I imagine we'll probably want something similar for the primary source sets application as well. At some point we should probably see if there's anything we could refactor into a `common_rails` role, but I don't think we need to do that immediately.

markbreedlove:
Thanks, @anarchivist.  I've taken care of that in the primary source sets branch:  https://github.com/dpla/automation/blob/feature/add-primary-source-sets/ansible/roles/pss/tasks/deploy.yml#L118-L124

Ultimately, I think that the correct way to get rid of the duplication throughout the playbooks is to write an Ansible module for these things.

PR #118, opened 2015-09-08
Only add cms group to Postgres when ingestion2 is not defined (fixes #117)

### LINKED ISSUE #117: Ingestion2 process install error
I'm following the [README-ingestion2](https://github.com/dpla/automation/blob/develop/README-ingestion2.md) document, and everything seems to work fine until I get to this point:

```
ansible-playbook -i ingestion -u <username> dev_ingestion_all.yml --extra-vars "initial_run=true"`
```

Most of the playbook seems to work, but it errors out partway through with:

```
TASK: [postgresql | Ensure state of PostgreSQL networking config file (pg_hba.conf)] *** 
fatal: [dev1] => {'msg': "AnsibleUndefinedVariable: One or more undefined variables: 'dict object' has no attribute 'cms'", 'failed': True}
fatal: [dev1] => {'msg': "AnsibleUndefinedVariable: One or more undefined variables: 'dict object' has no attribute 'cms'", 'failed': True}

```

Doing  a `vagrant reload` and re-provisioning keeps erring at the same point.  I'm unfortunately new to Ansible, and unfamiliar with the process, so I'm at a bit of a loss as to where to look to troubleshoot this further.  Any help would be greatly appreciated.

anarchivist:


ansible/roles/postgresql/templates/pg_hba.conf.j2
```diff
@@ -79,10 +79,12 @@ host   all             all            {{ host }}             md5
 {% for host in groups['api'] %}
 host   all             all            {{ host }}             md5
 {% endfor %}
+{% if ingestion2 is undefined %}
```
markbreedlove: is `undefined` correct?  I'm accustomed to `not defined`.  Otherwise, this looks ready.

markbreedlove: yeah, it's fine, acc. to http://jinja.pocoo.org/docs/dev/templates/

markbreedlove:
I really should have noticed this before, so sorry for the hassle, but the conditional for the `cms` group should be like the others, where the group's existence is verified with `if groups['cms'] is defined`. As the commit message states, in addition to the added consistency, this also has the potential to make the PostgreSQL configuration more flexible with role-specific inventory files that we might want to create
in the future.

markbreedlove:
Futrher note: I have tested this with both the `development` and `ingestion` inventories.

PR #119, opened 2015-09-11
[siteproxy] cms-group config only set when group is defined (fixes #117)

### LINKED ISSUE #117: Ingestion2 process install error

anarchivist:


markbreedlove:
That looks ready to go.

@anarchivist: have you checked it with a local VM, with the `ingestion` inventory?

anarchivist:
Yep, I've confirmed it with the local VMs in `ingestion`.

PR #120, opened 2015-09-14
Fix ruby bundle installations and upgrades

markbreedlove:
See https://issues.dp.la/issues/7840

This fixes issues with newly-added gems not installing, but ensures that deployments install the latest bugfix releases of correctly-pinned gems.


AudreyAltman:
:+1: This looks good to me.

no-reply:
:rocket:

PR #121, opened 2015-09-18
Improve PostgreSQL configuration for write performance

markbreedlove:
Add variables for some PostgreSQL configuration settings -- and adjust others -- that govern write performance. Change the configuration of the `bgwriter`, checkpointing, and locking. Enable logging for locks, checkpoints, and autovacuuming.

This has been tested with our `development` and `ingestion` inventories to the extent of making sure the playbook runs successfully. The settings have been used in our staging and production environments, with overrides for certain variables.

### DIFF
--- a/ansible/roles/postgresql/defaults/main.yml
+++ b/ansible/roles/postgresql/defaults/main.yml
@@ -1,6 +1,7 @@
 ---
 
 postgresql_user: {name: "dpla", password: "changeme"}
+# These memory-related values are tuned to the size of the development VM.
 pg_max_connections: 50
 pg_shared_buffers: 24MB
 pg_effective_cache_size: 128MB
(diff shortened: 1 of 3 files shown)


ansible/roles/postgresql/defaults/main.yml
```diff
@@ -9,6 +9,14 @@ pg_maintenance_work_mem: 16MB
 # The default random_page_cost of 4.0 assumes you're using disks.
 # Decrease it if you're using SSD drives.  1.5 might be a good value.
 pg_random_page_cost: 4.0
+# The following bgwriter parameters are PostgreSQL defaults:
```
anarchivist: Since you changed these in 4fc2f9002e17214e91bbed8fc1e192eac50530e8, should we remove this comment? Should we instead reference the defaults in comments?

anarchivist:
I'm :+1: once my comment gets resolved.

ansible/roles/postgresql/defaults/main.yml
markbreedlove: Yes, the comment is wrong.  Thanks for catching that.
I'll just put in updated comments explaining, where it's important, how these differ from the defaults.

markbreedlove:
OK, fixed those comments.
I should squash this before it gets merged.

PR #122, opened 2015-09-30
Move variables

markbreedlove:
This should fix https://issues.dp.la/issues/7888

The solution turned out to be more complex that what I imagined in that ticket, but results in a much easier to comprehend set of files, making it more obvious where a given variable is defined.

The basic idea is to rearrange our variables to fix some problems with overriding them, and to make their management easier.  It should be more clear now where a variable is defined and overridden, and there should be far fewer files to keep track of.

Upgrade instructions are in `README-upgrade-2.0.txt`.  Review of this PR should include confirmation that those instructions work.

These changes are targeted for a Version 2 release.  They will break any version 1 configuration because some variables have been moved and renamed.

The best way to review this is not necessarily to view the diff (unless you want to; 149 files have been changed), but to do a checkout and follow the upgrade instructions, especially running the suggested test, a dry run with the `-C` and `-D` options.  That should not report any significant changes.


anarchivist:
I followed your suggestions with the review, and checked out the code locally. I think it looks largely OK after I did the dry-run. However, it does look like we still need to fix the references to variable files in all of the  deployment/maintenance related playbooks (e.g. `ansible/playbooks/deploy_api_*.yml`).

markbreedlove:
Oh, right, thanks for flagging that. I forgot to reimplement those with tags, where the playbook specifies the role with a filter on the `deployment` tag.

markbreedlove:
That last commit should take care of the deployment playbooks.  I've finally had the chance to test running all of them and I don't see any errors, or any indications of unwanted changes.  Do you want to do any testing?

anarchivist:
When running `dev_ingestion_all.yml` in check/dry-run mode, I get an error:

```bash
halfsour:ansible mark$ ansible-playbook  -u mark -i ingestion -C -D dev_ingestion_all.yml

[... SNIP ...]

TASK: [frontend | Ensure existence of necessary packages] *********************
fatal: [dev1] => error while evaluating conditional: not skip_configuration

PLAY RECAP ********************************************************************
           to retry, use: --limit @/Users/mark/dev_ingestion_all.retry

dev1                       : ok=113  changed=9    unreachable=1    failed=0
dev2                       : ok=18   changed=0    unreachable=0    failed=0
```

markbreedlove:
Oops ... I'll check ...

markbreedlove:
@anarchivist: commit https://github.com/dpla/automation/commit/be4431679382f5f72705544fcbeb41a0a5dc1bb4 should fix that.

anarchivist:
OK, this looks good. I'm running into an issue w/ my dev VMs under the `ingestion` inventory/Vagrantfile where `dev2` (which runs Tomcat) ran out of memory to start up a JVM. I think that's a red herring here, but it's probably worth mentioning...

anarchivist:
Next steps: 

# Do we want to squash?
# Do we want to roll this out before Whiting launches?

markbreedlove:
Yes, I've encountered the memory situation, and it's something that I've deferred solving until later. It predates the changes in this PR.

I am in favor of squashing this tomorrow morning, and it would ideally be merged before Whiting launches, because changes I've made for Primary Source sets are branched from this branch.  Having it all merged by Monday would ensure easier deployments by people other than me.

anarchivist:
OK, that sounds fine to me - I just wanted to check in about it.

markbreedlove:
Squashed!

PR #125, opened 2015-10-19
Add settings and variables to pss settings.local.yml

markbreedlove:
* Add pss_contact_email variable for pss role
* Add Wordpress and Exhibitions settings


AudreyAltman:
:+1:

PR #126, opened 2015-10-19
Set RAILS_ENV when running Rails 4.x build scripts

anarchivist:
This addresses issues where assets were not being loaded, e.g. for the
primary-source-sets application and Heidrun.


markbreedlove:
Great!  Just for my information, does this assume that the pss app is using the fork of lightbox2-rails that uses assets-url instead of image_url?

This looks like something that we might want to do for our other Rails build scripts, too.  It would be OK if that were to come later.

I'm ready to merge this if there will be no further changes for this PR.

anarchivist:
No, this will work with the current release of `primary-source-sets`, and with `master` or `develop`. I've added the changes to the ingestion app build script too.

Basically, the issue is that without `RAILS_ENV` being explicitly passed, it's treating it as `development`.

markbreedlove:
This also fixed the fact that the CSS and JS files had been getting concatenated, but not minified.  They're now minified.

PR #127, opened 2015-10-20
Add pss to roles that have turnout maintenance mode

markbreedlove:
Add the Primary Source Sets app to those Rails apps that are connected to PostgreSQL and that need to be taken offline during PostgreSQL maintenance.


anarchivist:
:rocket:

PR #129, opened 2015-10-26
Fix variable name

markbreedlove:
Change `pss_rails_env` to `ingestion_app_rails_env`

Causes an error due to the variable not being defined when run with defaults (without it being overridden).


anarchivist:
:+1:

PR #130, opened 2015-10-28
Improve installation instructions

markbreedlove:
* Add step for configuring locale before running subsequent package  installations.  Especially necessary to get the right encodings of  PostgreSQL's system databases.
* Add instruction in Ingestion 2 Readme to download Ubuntu 14.04 LTS box  in addition to 12.04.

The locale and encoding instructions are important to have in-place before merging some upcoming changes that perform a PostgreSQL 9.4 upgrade.

### DIFF
--- a/README-ingestion2.md
+++ b/README-ingestion2.md
@@ -13,10 +13,11 @@ Follow all of the steps in [README.md](README.md) up until "Copy Vagrantfile.dis
 
 * Copy `Vagrantfile.ingestion2` to `Vagrantfile`
 
-* Make sure that Vagrant has downloaded the base server image that we'll need
+* Make sure that Vagrant has downloaded the base server images that we'll need
   for our VMs:
 ```
 $ vagrant box add hashicorp/precise64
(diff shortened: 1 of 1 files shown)


markbreedlove:
Actually, *please hold off merging this*.  The locale gets established correctly, and picked up by the PostgreSQL installation, with the legacy `development` inventory and `dev_all.yml` playbook, but not with the `ingestion` inventory and `dev_ingestion_all.yml` playbook.  That is what actually needs to be changed.

The line about adding `ubuntu/trusty64` is still valid.

anarchivist:
Sounds good; just let me know when it's ready.

markbreedlove:
OK, it's ready now.  After doing a number of tests, I can not figure out what accounts for the difference between legacy and Ingestion 2, and I just have to move on with the PostgreSQL upgrade.

The instructions as they exist now should work.  It's only in the Ingestion 2 Readme that it says to perform the locale configuration with its own command, and this works for now, until I can figure out something better.

no-reply:
:+1:

PR #132, opened 2015-10-30
Install frontend assets only if requested

markbreedlove:
Add the variable `frontend_include_branding`, which governs whether the private dpla_frontend_assets gem is included for branding-related assets.  This defuaults to false.

Addresses one of the problems raised in https://github.com/dpla/automation/issues/117

There's a known issue mentioned in the comments of `build_frontend.sh`, but this is as good as it gets for now.  This prevents non-DPLA staff from running into build errors.

### DIFF
--- a/ansible/roles/frontend/files/build_frontend.sh
+++ b/ansible/roles/frontend/files/build_frontend.sh
@@ -1,34 +1,55 @@
 #!/bin/bash
 
+# Synopsis
+#
+# - Build with Ruby 1.9.3-p547, with dpla_frontend_assets gem
+#   build_frontend.sh 1.9.3-p547
+#
+# - Build as above, but with dpla_frontend_assets gem
+#   build_frontend.sh 1.9.3-p547 branding
+
+
 USE_VERSION=$1
+BRANDING=${2:-""}
 export PATH=$HOME/.rbenv/bin:$PATH
 LOGFILE=/tmp/build_frontend.log
 
 echo "starting" > $LOGFILE
 
 eval "`rbenv init -`" >> $LOGFILE 2>&1
-# Start ssh-agent and set environment variables.
-# Work-around for private GitHub repository in Gemfile.
-eval `ssh-agent`
-ssh-add $HOME/git_private_key  >> $LOGFILE 2>&1
+
+echo "rbenv initialized" >> $LOGFILE
+
+if [ "$BRANDING" == "branding" ]; then
+    # Start ssh-agent and set environment variables.
+    # Work-around for private GitHub repository in Gemfile.
+    echo "starting ssh-agent ..." >> $LOGFILE
+    eval `ssh-agent`
+    ssh-add $HOME/git_private_key  >> $LOGFILE 2>&1
+fi
 
 cd $HOME/frontend
 
 rbenv shell $USE_VERSION >> $LOGFILE 2>&1
+echo "using rbenv version $USE_VERSION" >> $LOGFILE
 
 echo "installing bundle ..." >> $LOGFILE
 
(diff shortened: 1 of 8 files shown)


ansible/roles/frontend/files/build_frontend.sh
```diff
@@ -19,16 +47,26 @@ rbenv shell $USE_VERSION >> $LOGFILE 2>&1
 echo "installing bundle ..." >> $LOGFILE
 
 rm -f Gemfile.lock
-bundle install >> $LOGFILE 2>&1
-bundle update dpla_frontend_assets >> $LOGFILE 2>&1
+if nobranding; then
+    install_opts="--without dpla_branding"
+else
+    install_opts="--with dpla_branding"
```
anarchivist: `--with` is not a Bundler parameter, to the best of my knowledge. See http://bundler.io/bundle_install.html; this probably means `install_opts` should just be blank.

anarchivist: Never mind; it actually is a parameter, it's just undocumented.

anarchivist:
I'll try this out this weekend, but my sense is 312623f should not be necessary.

markbreedlove:
There might be some other way but `--without` wasn't working (Bundler kept fetching the gem) and deleting the line from the file is what finally gave me some relief.

I was testing this with `alias ansible-playbook='ansible-playbook -e "github_private_key_path=~/.ssh/id_rsa_test"'`, where id_rsa_test is a key that does not have access to the private repo.

markbreedlove:
I checked in with @AudreyAltman earlier this morning and she confirmed that, based on a past experience of hers, the only solution is to remove the gem from the Gemfile in deployment.  She said she came across a discussion in a forum or list debating the correctness of this behavior, but that Bundler will in fact not forego fetching the gem if it's in the Gemfile.

anarchivist:
In that case, I have a proposal - how about if we remove the restricted assets gem from the `Gemfile` and add it conditionally? I'd rather make that kind of surgical change in deployment than add to the complexity here.

markbreedlove:
OK, I'll see about submitting a quick PR for `frontend`, assuming that will be quick.

markbreedlove:
Bu wait ... don't we still add complexity here to add it to the gemfile in `automation`?

no-reply:
> In that case, I have a proposal - how about if we remove the restricted assets gem from the Gemfile and add it conditionally?

:+1: to this.

Slack discussion proceeds from https://dpla.slack.com/archives/tech/p1446485988000043.  I agree with @markbreedlove that this is probably just a bad way of handling the asset deployment. Short of fixing that larger issue, @anarchivist's proposal seems best.

markbreedlove:
fe5b97d reverses the strategy to add the gem to the Gemfile if branding is turned on.

This depends upon an upcoming PR to `frontend` that removes the gem.

I've tested this by deploying it in development and can switch back and forth between branded and unbranded deployments.

anarchivist:
While I'm thinking of it, do we need to do the same thing with the `primary-source-sets` application?

markbreedlove:
I think so ... cc: @AudreyAltman

markbreedlove:
I'm going to make a similar change for this PR that does the same trick with the primary-source-sets Gemfile and assumes that the primary-source-sets app has an ensuing PR to remove the gem.

markbreedlove:
I think this should be good now, based on my testing.  We should have the related changes to `frontend` and `primary-source-sets` merged before this goes public.

anarchivist:
It looks like this needs to be rebased on `develop`.

PR #134, opened 2015-10-30
Allow setting of non-inventory database host

markbreedlove:
For various roles, add variables that allow for database hosts that exist outside of the inventory.  A good example is an Amazon RDS database instance that's not used in development.

The variables default to false so that normally the PostgreSQL database host is used.

To verify the behavior, you should be able to run `ansible-playbook` with `-C -D` and observe no changes to the `database.yml` config files, except for the removal of one line that I sneaked in there to remove a bogus property `hostname` from the `ingestion_app` file.

If you override something like `api_alternate_db_host` you should be able to see a change being proposed to the `database.yml` file with `-C -D`.

### DIFF
--- a/ansible/playbooks/pg_maintenance_off.yml
+++ b/ansible/playbooks/pg_maintenance_off.yml
@@ -8,6 +8,7 @@
       file: >-
         path=/srv/www/api/tmp/maintenance.yml
         state=absent
+      when: not api_alternate_db_host | default(false)
 
 - name: Take frontend out of maintenance mode
   hosts: frontend
@@ -17,6 +18,7 @@
       file: >-
         path=/srv/www/frontend/tmp/maintenance.yml
         state=absent
+      when: not frontend_alternate_db_host | default(false)
 
 - name: Take Primary Source Sets out of maintenance mode
   hosts: pss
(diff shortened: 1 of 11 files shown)


markbreedlove:
Assigning to @AudreyAltman to spread out the load of assigned PR reviews!  Feel free to bump to @anarchivist or @no-reply if necessary.

anarchivist:
:+1:

markbreedlove:
I need to add some conditionals to avoid putting apps into maintenance mode that will not need it anymore if the PostgreSQL database or server are restarted.

AudreyAltman:
I tested this locally and it worked as expected.  @markbreedlove when you've had a chance to add the conditionals, this looks good.

markbreedlove:
Thanks, @AudreyAltman.  I am working on the conditionals which are not as simple as I thought due to an annoyance with `where` in the context of an `include` that will be fixed when Ansible 2.0 is released.

markbreedlove:
Hold off ... that last commit needs improvement.

markbreedlove:
OK, there you go.  @anarchivist and @AudreyAltman, I think this is it.

anarchivist:
:white_check_mark: :jack_o_lantern: :question:

AudreyAltman:
:+1:

PR #135, opened 2015-11-02
Set default port for PSS application

anarchivist:
A new from-scratch deployment fails if you're just copying `group_vars/all.dist` to `group_vars/all`. There is no default value for the `pss_app_port` variable.


markbreedlove:
Thanks!

markbreedlove:
I'll merge this after https://github.com/dpla/automation/pull/134 because I just squashed and rebased that one.

PR #137, opened 2015-11-05
Fix VM memory overallocation

markbreedlove:
Reduce the amount of memory taken by Ruby processes in the 'ingestion' inventory for Ingestion 2 by reducing the number of processes.

Also give a little more memory to the 'dev2' VM in the Vagrantfile.

The worker processes could be dropped to just one, but I figure that it could be useful in development to make sure that the system works with concurrently-running processes.


anarchivist:
:+1:

PR #138, opened 2015-11-10
Make twofishes start with system locale

markbreedlove:
Fix the init script to export the LANG locale environment variable, which fixes an error with the startup of twofishes, where it is not happy trying to read its database with the ASCII encoding.  It should try to use UTF8.

The `twofishes` init script was failing to start `twofishes` upon reboot.


no-reply:
Looks pretty simple. :+1:

anarchivist:
:white_check_mark:

PR #139, opened 2015-11-12
Disable Spring wherever we use Ruby

markbreedlove:
Add `DISABLE_SPRING=1` to the `dpla` user's `.bashrc` wherever we set up Ruby.  We've always had problems running the Rails console whenever Spring has been running.  This is mostly a development issue.

I chose `install_ruby_tools.sh` because that's where other Ruby-related things are already set up, like `rbenv` and `bundler`.


no-reply:
This! Big :+1:.

:rocket:

PR #140, opened 2015-12-01
Add new Bulk Download file script

markbreedlove:
This script generates and uploads Bulk Download files straight from the Elasticsearch index, and supercedes the file-generation functionality of the legacy [ingestion](https://github.com/dpla/ingestion) project.

This change includes the automation necessary to have it run automatically on the `central` server, if that exists.

The script is designed to be useful either on cron or as a tool for exporting individual files on command. As such, it can be run locally or on the server.

### DIFF
--- a/ansible/roles/exporter/defaults/main.yml
+++ b/ansible/roles/exporter/defaults/main.yml
@@ -1,0 +1,8 @@
+---
+
+exporter_access_key: changeme
+exporter_secret_key: changeme
+exporter_bucket: changeme
+exporter_cron_time: "30 0 1 * *"
+exporter_output_dir: /var/tmp
+exporter_s3_bucket: changeme
--- a/ansible/roles/exporter/files/export-provider.sh
+++ b/ansible/roles/exporter/files/export-provider.sh
@@ -1,0 +1,191 @@
+#!/bin/bash
+
+## export-provider.sh
+#
+#  Export the DPLA search index to files, per-provider or for all items, and
+#  upload these files to an Amazon S3 bucket.
+#
+#  These files are known as Bulk Download files:
+#  http://dp.la/info/developers/download/
+
+SCRIPT=`basename ${BASH_SOURCE[0]}`
+
+which aws > /dev/null && which elasticdump > /dev/null && which jq > /dev/null
+if [ $? -ne 0 ]; then
+    >&2 echo "\`aws', \`elasticdump', and \`jq' must be available in \$PATH."
+    exit 1
+fi
+
+# Figure out `sed' options for extended regexes
+sed -r 2>/dev/null
+if [ $? -eq 0 ]; then
+    sed_opts="-r"  # GNU
+else
+    sed_opts="-E"  # OS X
+fi
+
+
+usage() {
+    cat <<END
+
+Usage:  $SCRIPT -u <url> -o <directory> -s <s3 bucket> -p <provider>...
+
+    -u    Elasticsearch URL (including index)
+    -o    Output directory
+    -s    S3 destination bucket
+    -h    Help (this message)
+    -p    The name of one provider, or "all" for all of them, or "each"
+          for each one (queried from the search index).  "all" and
+          "each" can be used together.  Multiple providers may be specified.
+          The provider name corresponds to provider.@id in our schema.
+
+END
+}
+
+# Generate the Elasticsearch Query API syntax that is used to select which
+# records to export
+searchbody() {
+    if [ "$@" == "all" ]; then
+        echo '{"query": {"match_all": {}}}'
+    else
+        cat <<END
+{
+    "query": {
+        "query_string": {
+            "default_field": "provider.@id",
+            "query": "*$1"
+        }
+    }
+}
+END
+    fi
+}
+
+# Generate the Elasticsearch Query API syntax that is used to get a list of
+# all providers by provider.@id
+providers_list_search_body() {
+    cat <<END
+{
+    "query": {
+        "match_all": {}
+    },
+    "facets": {
+        "providers": {
+            "terms": {
+                "field": "provider.@id",
+                "size": 500
+            }
+        }
+    },
+    "size": 0
+}
+END
+}
+
+
+providers_list_response() {
+    curl -s -X POST "$ES_URL/_search" \
+        -d "`providers_list_search_body`"
+    if [ $? -ne 0 ]; then
+        >&2 echo "Could not query provider listing"
+        exit 1
+    fi
+}
+
+# Given JSON produced by `providers_list_search_body`, return a list of tokens
+# that represent the substrings at the end of provider.@id.
+# For example, "nypl" for "http://dp.la/api/contributor/nypl"
+providers_list() {
+    json=`providers_list_response`
+    echo $json | jq '.facets.providers.terms[].term' \
+        | sed $sed_opts 's/^.*\/([^\/]+)"$/\1/'
+}
+
+# Return basename of the output file, which is gzipped JSON
+outfile() {
+    base=`echo $1 | sed $sed_opts 's/[^A-Za-z]+/_/g;'`
+    echo ${base}.json.gz
+}
+
+# Export a provider (or "all" providers) and upload the file
+export_and_upload() {
+    sbody=`searchbody "$@"`
+    ofile=`outfile "$@"`
+    output="${OUTDIR}/${ofile}"
+    elasticdump --searchBody "$sbody" \
+        --input=$ES_URL --output=$ | gzip > $output  \
+        || exit 1
+    bn=`basename $output`
+    dir=`date +"%Y/%m"`
+    aws s3 cp --quiet  \
+        --content-type "application/gzip" --content-encoding "identity" \
+        $output s3://$BUCKET/$dir/$bn \
+        || exit 1
+    rm $output || exit 1
+}
+
+declare -a providers
+provider_idx=0
+# Get options
+while getopts :f:u:o:s:p:h flag; do
+    case $flag in
+        u)
+            ES_URL=$OPTARG
+            ;;
+        o)
+            OUTDIR=$OPTARG
+            ;;
+        s)
+            BUCKET=$OPTARG
+            ;;
+        p)
+            providers[$provider_idx]="$OPTARG"
+            provider_idx=$(( $provider_idx + 1 ))
+            ;;
+        h)
+            usage
+            exit 0
+            ;;
+        \?)
+            usage
+            exit 1
+            ;;
+    esac
+    i=$(( $i + 1 ))
+done
+shift $((OPTIND-1))  # Move to next argument
+
+# Check various arguments
+#
+if [ "x$ES_URL" == "x" ]; then
+    >&2 echo "Unknown Elasticsearch URL"
+    usage
+    exit 1
+fi
+
+if [ "x$OUTDIR" == "x" ]; then
+    >&2 echo "Unknown output directory"
+    usage
+    exit 1
+fi
+OUTDIR=`echo $OUTDIR | sed 's/\/$//'`
+
+if [ "x$BUCKET" == "x" ]; then
+    >&2 echo "Unknown S3 bucket"
+    usage
+    exit 1
+fi
+
+## Export each provider
+#
+for (( j=0; $j < $provider_idx ; j=$(( j + 1 )) )); do
+    if [ "${providers[$j]}" == "each" ]; then
+        for provider in `providers_list`; do
+            export_and_upload "$provider"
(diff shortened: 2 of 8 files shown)


markbreedlove:
Hold on -- there are some issues I need to fix ...

markbreedlove:
OK, should be good to review now.

ansible/roles/exporter/files/export-provider.sh
```diff
@@ -0,0 +1,169 @@
...
+#
+#  Export the DPLA search index to files, per-provider or for all items, and
+#  upload these files to an Amazon S3 bucket.
+#
+#  These files are known as Bulk Download files:
+#  http://dp.la/info/developers/download/
+
+SCRIPT=`basename ${BASH_SOURCE[0]}`
+ELASTICDUMP=${ELASTICDUMP:-/usr/local/bin/elasticdump}
+AWS=${AWS:-/usr/local/bin/aws}
```
markbreedlove: If anything needs to be changed, it's the location of the `elasticdump` and `aws` executables. It may be better to set `PATH` to include `/usr/local` in the crontab, if that is necessary. With it working the way it is now, you have to go through the extra hassle of defining `AWS` if you installed it in a Python virtualenv, or `ELASTICDUMP` if you installed it locally with `npm`. This script works the way it does now to make it run easily on cron, but it probably provides more clarity to simply configure the path as necessary in the crontab file.

anarchivist: Alternately, I'm wondering if could these be symlinked somewhere predictable I don't have a strong sense of what's preferable, and whether `aws` or `elasticdump` can access their dependencies successfully on their own.

anarchivist: (Never mind the previous comment. I was just commenting absentmindedly, and forgot that this script was intended to invoke both executables.)

ansible/roles/exporter/files/export-provider.sh
```diff
@@ -0,0 +1,169 @@
...
+# records to export
+searchbody() {
+    if [ "$@" == "all" ]; then
+        echo '{"query": {"match_all": {}}}'
+    else
+        cat <<END
+{
+    "query": {
+        "query_string": {
+            "default_field": "provider.name",
```
anarchivist: I'm wondering if it makes more sense to use the `provider.@id`, which is more likely to (and should) remain consistent over time. (For example, we changed the `skos:prefLabel` from "U.S. Government Printing Office" to "U.S. Government Publishing Office" somewhat recently.)

ansible/roles/exporter/files/export-provider.sh
```diff
@@ -0,0 +1,169 @@
...
+        }
+    }
+}
+END
+    fi
+}
+
+# Return basename of the output file, which is gzipped JSON
+outfile() {
+    base=`echo $1 | sed $sed_opts 's/[^A-Za-z]+/_/g;' | tr '[:upper:]' '[:lower:]'`
```
anarchivist: Ah, I see that this might be an issue then if we use `provider.@id`.

ansible/roles/exporter/files/export-provider.sh
```diff
@@ -0,0 +1,169 @@
...
+# Export a provider (or "all" providers) and upload the file
+export_and_upload() {
+    sbody=`searchbody "$@"`
+    ofile=`outfile "$@"`
+    output="${OUTDIR}/${ofile}"
+    $ELASTICDUMP --searchBody "$sbody" \
+        --input=$ES_URL --output=$ | gzip > $output  \
+        || exit 1
+    bn=`basename $output`
+    $AWS s3 cp --quiet --content-encoding "gzip" $output s3://$BUCKET/$bn \
```
anarchivist: Would it be possible to specify year and month as part of the key for the file added to the S3 bucket (e.g. `s3://$BUCKET/2015/12/internet_archive.json.gz`)?

anarchivist:
I think this looks good, but I'd like to discuss the impact of `provider.name` vs `provider.@id`.

ansible/roles/exporter/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
+---
+
+exporter_access_key: changeme
+exporter_secret_key: changeme
+exporter_bucket: changeme
+exporter_cron_time: "30 0 1 * *"
+exporter_output_dir: /var/tmp
+exporter_s3_bucket: changeme
+exporter_providers:
```
anarchivist: I'm also wondering if `exporter_providers` can be populated automatically some way, pulling down the values for the `provider.@id`/`provider.name` facet. That might be worth following up on later, though.

markbreedlove:
Just checked, and it should be easy enough to query `provider.@id` from the script itself, and to generate file names based on this string.

markbreedlove:
Regarding the executable paths:  I think now that the best thing is to let them be found in `PATH` and have it be the cron job's responsibility to have that set.

markbreedlove:
@anarchivist: I've made some changes related to the comments above.
* Providers are now selected by `provider.@id` instead of `.name`. Files are generated with the string at the end of `@id`.
* There's a feature to automatically query the list of providers and generate files for all of them, which takes care of new providers being added.
* Bucket file names have "folders" like "2015/12"

anarchivist:
:+1: passing to @no-reply for a second look

ansible/roles/exporter/files/export-provider.sh
markbreedlove: With the latest commits, the translation of uppercase to lowercase is no longer necessary, but it could still be nice to normalize "-" and "_" to "_".  The `@id`s are inconsistent and use one or the other.

ansible/roles/exporter/files/export-provider.sh
```diff
@@ -0,0 +1,188 @@
...
+export_and_upload() {
+    sbody=`searchbody "$@"`
+    ofile=`outfile "$@"`
+    output="${OUTDIR}/${ofile}"
+    elasticdump --searchBody "$sbody" \
+        --input=$ES_URL --output=$ | gzip > $output  \
+        || exit 1
+    bn=`basename $output`
+    dir=`date +"%Y/%m"`
+    aws s3 cp --quiet --content-encoding "gzip" $output s3://$BUCKET/$dir/$bn \
```
markbreedlove: This is causing a file to be served with `Content-Type: application/json` and `Content-Encoding: gzip`, which I think is the recommended way to do it, but makes a browser display the file instead of saving it. The user has to deliberately click "save as." Is everyone OK with this? This is not an issue for someone using a script, or using curl / whatever.

AudreyAltman: FWIW I think that as long as it's clear to users how to save the file, it doesn't matter that it isn't automatically saved.

anarchivist: @AudreyAltman @markbreedlove The major issue here is that the gzip-compressed JSON blobs are highly compressed. For example, `artstor.json.gz` currently is 25 MB when compressed, but is 363 MB uncompressed. With  `Content-Type: application/json` and `Content-Encoding: gzip`, this causes the browser to both uncompress AND display the contents of the file. 

Recommendation, per [this Stack Overflow answer](http://stackoverflow.com/a/6007788/1277323): Set headers as  `Content-Type: application/octet-stream` and `Content-Disposition: attachment` (setting the filename is not necessary). I'm guessing that we *don't* want to specify `Content-Encoding: gzip` here as well; therefore, we'd address all this by changing line 119 to:

```
aws s3 cp --quiet --content-type "application/octet-stream" --content-disposition "attachment" $output s3://$BUCKET/$dir/$bn \
```

ansible/roles/exporter/files/export-provider.sh
no-reply: > With the latest commits, the translation of uppercase to lowercase is no longer necessary...

Do we want to remove the `tr` call?

markbreedlove: Yes, I'll remove it, but I'll normalize the "-" and "_" characters.

ansible/roles/exporter/files/export-provider.sh
markbreedlove: I'll try that ...

ansible/roles/exporter/files/export-provider.sh
markbreedlove: Never mind about normalization; already being done, of course. I'll just remove the `tr`.

ansible/roles/exporter/files/export-provider.sh
no-reply: `application/octet-stream` seems to work for browsers, but I'm wondering about how it might affect other agents. Do we expect people to be pulling bulk downloads with other software that might be relying on the `Content-Type`?

I'm inclined to see things like this [other Stack Overflow user](http://stackoverflow.com/a/10616753).  Hacking user agent behavior with misleading headers seems fragile.

no-reply:
With the exception of the outstanding `Content-Type` discussion, I'm :+1: on this.

I'm going to generate some files and parse them as JSON-LD, before merging. Hopefully that should be done around the same time as the issues here are resolved. :)

ansible/roles/exporter/files/export-provider.sh
markbreedlove: I agree on not lying about the content type.  I'm seeing if content-type attachment works.

anarchivist: Per that post, the `Content-Disposition` header is not misleading, it's the recommended behavior.

If I'm not mistaken, if we specify `Content-Type: application/json` and `Content-Encoding: gzip` the client will automatically decompress the resource. I think this is the most thing to avoid at this point.

no-reply: @anarchivist I'm :+1: on `Content-Disposition: attachment`; but (tentatively) :-1: on `Content-Type:  application/octet-stream`.

> If I'm not mistaken, if we specify Content-Type: application/json and Content-Encoding: gzip the client will automatically decompress the resource. I think this is the most important thing to avoid at this point.

Don't current browsers respect the `Content-Disposition` header?  If not, I think I agree that we should be willing to underspecify the content-type to avoid bad behavior. 

Generally, I would prefer to count on the user agent to make good decisions about how to react to the files; but if we have to hack, we have to hack. :disappointed:

anarchivist: > Don't current browsers respect the `Content-Disposition` header? If not, I think I agree that we should be willing to underspecify the content-type to avoid bad behavior.

Yes, they do, but they still uncompress gzipped streams if `Content-Encoding: gzip` is specified (I tested this in Chrome last week).

no-reply: Ah. [This answer](http://superuser.com/a/942580) is also useful. Chrom(ium) interprets the `Content-Encoding: gzip` header as meaning that the compression is used to save bandwidth; that we want the user to receive JSON.

I'm on board now. :+1: to @anarchivist's proposal.

Thanks for your patience, @anarchivist. :)

no-reply: We could also consider `application/gzip` as the content type. Thoughts?

anarchivist: > We could also consider `application/gzip` as the content type. Thoughts?

:+1:

markbreedlove: I was lazy in my responses. To explain further:

My original understanding was that it was correct to set Content-Type to application/javascript and Content-Encoding to gzip.  Depending on the conditions given in the Accept-Encoding header, this is often the case for a JSON response in a typical API -- but not what we want in our case.

I shouldn't have said OK so quickly to application/octet-stream; that's for files that are binaries after all decoding is applied. Content-Type is supposed to represent the content type that the client has to work with after the decoding of the kind specified in Content-Encoding is done.

In our case, I think it's fair to say that the user really wants a gzip file and doesn't expect to have an uncompressed JSON file. This doesn't have anything to do with an Accept-Encoding as part of an API request; we really want to deliver a gzipped file, not a huge uncompressed one, even though the underlying content is JSON.

I can confirm that it works to have `Content-Type: application/json` and  `Content-Disposition: attachment`. You download the file when you click on it in a browser, and end up with a `.gz`. However, I think it would actually be best to have `Content-Type: application/gzip` and `Content-Encoding: identity`. I can confirm that this works, too, and I think it better represents the fact that the client is supposed to end up with a .gz file that needs further handling.

no-reply: Sounds like we're all on the same page. :+1: to `application/gzip`

no-reply:
Awesome!

PR #141, opened 2015-12-10
Fix undefined variable, elasticsearch reboot

markbreedlove:
Change the invocation of a local action to run the set-allocation.sh script in order for the es_cluster_loadbal variable to be inherited from the relevant group_vars file.

This fixes an error about that variable being undefined. This issue was introduced back when the variables were rearranged.


anarchivist:
:+1:

PR #142, opened 2015-12-14
Feature/heidrun mappings rsync rebased

no-reply:


no-reply:
This is a rebased version of #108.

anarchivist:
Closed by #108.

PR #143, opened 2016-01-13
add pss tags settings

AudreyAltman:
Add tags config setting to primary source sets.

### DIFF
--- a/ansible/roles/pss/defaults/main.yml
+++ b/ansible/roles/pss/defaults/main.yml
@@ -27,3 +27,4 @@ pss_zencoder_s3_credentials_name: false
 pss_contact_email: email@example.com
 pss_alternate_db_host: false
 pss_include_branding: false
+pss_display_tags_on_public_ui: false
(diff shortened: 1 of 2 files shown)


AudreyAltman:
This compliments [aws #31](https://github.com/dpla/aws/pull/31).

ansible/roles/pss/defaults/main.yml
```diff
@@ -27,3 +27,4 @@ pss_zencoder_s3_credentials_name: false
 pss_contact_email: email@example.com
 pss_alternate_db_host: false
 pss_include_branding: false
+pss_display_tags_on_public_ui: OVERRIDE-THIS
```
markbreedlove: This should probably be whatever default exists in `settings.yml` in the application.

I think that this string evaluates to true in a boolean context in Ruby.  I think that if  you experiment with it in your VM you should be able to determine how it gets written to settings.local.yml and then how that gets interpreted by the app.

AudreyAltman:
Okay, I've changed the default value to `false` to match what's in the `pss` app.  Also, I've been using boolean values in my local instance and they've been working fine, so we should be good with booleans (unless there's some reason why it would behave differently on staging/production).

markbreedlove:
OK, thanks!

PR #144, opened 2016-01-22
Add Primary Source Sets mail settings

markbreedlove:
See https://github.com/dpla/primary-source-sets/pull/104

There are some new Primary Source Sets application mail settings that need to be configured.

See http://guides.rubyonrails.org/action_mailer_basics.html#action-mailer-configuration and http://api.rubyonrails.org/classes/ActionMailer/Base.html

We set a default of `file` for development, because we don't run a MTA on the development VMs.  Staging and production environments can override this with `sendmail` so that `/usr/sbin/sendmail` is used.


markbreedlove:
With the recent merge of the account management feature into `primary-source-sets`, this will need to be released soon.

AudreyAltman:
:+1:

PR #145, opened 2016-01-26
Fix sqlite3 dependency, frontend app

markbreedlove:
Add these Ubuntu packages, because we've recently added the `sqlite3` gem to the `frontend` app for tests, and bundler wants to install it regardless of the environment.

### DIFF
--- a/ansible/roles/frontend/files/build_frontend.sh
+++ b/ansible/roles/frontend/files/build_frontend.sh
@@ -36,7 +36,7 @@ echo "using rbenv version $USE_VERSION" >> $LOGFILE
 echo "installing bundle ..." >> $LOGFILE
 
 rm -f Gemfile.lock
-bundle install >> $LOGFILE 2>&1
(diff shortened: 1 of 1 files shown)


anarchivist:
Forgive me, because we may have gone over this before, but could we solve this by changing the deployment script to call `bundle install --without test`?

markbreedlove:
Oh, I hadn't thought of that.  It sounds like a better approach.

markbreedlove:
@anarchivist: I added https://github.com/dpla/automation/commit/a58321d27766dbbbab995e6443b27afa844dc2d0 so that we're using `bundle install --without test` and I've seen it work on a system that has no sqlite3 or libsqlite3-dev packages.

anarchivist:
:+1: Squash?

markbreedlove:
Squished...

anarchivist:
:racehorse:

PR #146, opened 2016-01-27
Update API role for removal of contentqa engine

markbreedlove:
See https://github.com/dpla/platform/pull/40 and https://issues.dp.la/issues/8262

The `contentqa` engine is being removed from the platform (api) app, so we must remove the deployment rake tasks and service handlers pertaining to it.


anarchivist:
Should we clean up `roles/api/tasks/main.yml` to remove [lines 49-80](https://github.com/dpla/automation/blob/develop/ansible/roles/api/tasks/main.yml#L49-L80) and [lines 104-114](https://github.com/dpla/automation/blob/develop/ansible/roles/api/tasks/main.yml#L104-L114), and to add steps that remove the control and init scripts? 

Similarly, should we delete `delayed_job_api.sh.j2`, `etc_init.d_delayed_job_api.j2`,  and `qa_htpasswd.j2` from `roles/api/templates`?

markbreedlove:
Oh, yikes, yes.

markbreedlove:
Plus the Nginx block for `/qa` ...

markbreedlove:
Hang on, one more thing to add ...

markbreedlove:
@anarchivist: I think I'm done making my changes.

anarchivist:
OK, looks good to me. Adding a reference to the issue I created yesterday. https://issues.dp.la/issues/8262

Issue #148, opened 2016-01-29
Fix delayed_job cleanup ["pkill", "-9", "-f", "delayed_job"]


telezoic:
I've evened up my fork with the latest release, and re-installing on my local machine I get and error at he end of the invocations section in README-ingestion2.md:
 
```$ ansible-playbook -i ingestion -u <your username in group_vars/all> \
  dev_ingestion_all.yml --extra-vars "initial_run=true"```

The error: 
```TASK [api : Stop obsolete delayed_job if it is running] ************************```
```fatal: [dev1]: FAILED! => {"changed": true, "cmd": ["pkill", "-9", "-f", "delayed_job"], "delta": "0:00:00.003681", "end": "2016-01-29 13:10:56.166660" . . .  ```

I try a ```vagrant reload``` to no effect. I'm hoping you could point me in the right direction to solving this.

**Comment by markbreedlove** [comment]
Thanks for reporting this. I'll have a look soon.
Just to make sure I understand clearly, is this a new installation on a new VM?  (You mentioned re-installing, and you're running with `initial_run=true`.)

**Comment by telezoic** [comment]
Correct. New VM and with ```initial_run=true```. I'm going to come at this one more time in the interim.

**Comment by markbreedlove** [comment]
This is probably the problem.  If a process is not running, `pkill` says "Killed" but returns nonzero. That's getting interpreted (rightly) as failure by ansible.
```
mb@dev1:~$ ps auxw | grep [d]elayed
mb@dev1:~$ sudo pkill -9 -f delayed_job
Killed
mb@dev1:~$ echo $?
137
```

I think it'll be safe to have the task ignore errors.

telezoic:
I set it to ignore for now and all is well. Thanks for this!

PR #149, opened 2016-01-29
Fix pkill task to stop old delayed_job process

markbreedlove:
`pkill -f` actually returns an error status if the process did not exist, although it prints "Killed" to the terminal (not stdout or stderr -- I'm not sure which file it goes to).

Interestingly, plain `pkill` (without -f) does _not_ print "Killed" if the process did not exist, and it exits with 1.

The status that `pkill -f` returns if the process did not exist is 137, if a SIGKILL (-9) fails.  See http://tldp.org/LDP/abs/html/exitcodes.html 

The ansible `command` module was giving the return code as 1 in this case, instead of 137. The `shell` module, on the other hand, returns -9, which is not what we'd expect (137) but is better than 1, which is
indistinguishable from other kinds of failures.

### DIFF
--- a/ansible/roles/api/tasks/main.yml
+++ b/ansible/roles/api/tasks/main.yml
@@ -50,7 +50,12 @@
 # Clean up delayed_job init scripts and processes that used to be created in
 # earlier versions of this file
 - name: Stop obsolete delayed_job if it is running
-  command: pkill -9 -f delayed_job
+  shell: pkill -9 -f delayed_job
+  register: pkill_result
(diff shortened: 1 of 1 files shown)


markbreedlove:
Fixes https://github.com/dpla/automation/issues/148

markbreedlove:
Assigning to @anarchivist so that it's assigned to someone. Let me know if it should be bumped to someone else!

no-reply:
Looks good. I'm :+1: to merge.

PR #150, opened 2016-02-18
Add the svc-backup user to the postfix generic profile

moltude:


anarchivist:
Looks good. I'm imagining that the user will get added as part of an other pull request?

moltude:
yeah, I looked through the automation repo and wasn't sure how to do that.

markbreedlove:
Here's an example:
https://github.com/dpla/automation/blob/v2.5.1/ansible/roles/common_web/tasks/main.yml#L44-L50

You'll want to create a new Ansible role for the purpose of creating the user, installing the backup script, and putting it on a crontab. The `tasks/main.yml` playbook for that role would be the place to put the account-creation task.

markbreedlove:
We're using AWS Lambda for EBS snapshots now, so I'm closing this out.

PR #151, opened 2016-02-19
Allow alternate repos for ingestion

markbreedlove:
Parameterize the GitHub repositories that are used for `heidrun` and `heidrun_mappings`.  This allows us to do deployments into our test or development systems from forked repositories.


no-reply:
Seems reasonable.

It's a bit of a pain that these are on HM's fork; makes it harder to push changes, too.

PR #153, opened 2016-02-24
Add ingestion app Marmotta read_timeout setting

markbreedlove:
Add a parameter for the read_timeout setting that was recently added to Krikri.

See
https://github.com/dpla/KriKri/commit/c8a4c3596c95ef0800dc27f06421e3642ebe4754


moltude:
:+1: to me but maybe @no-reply wants to take a quick look

no-reply:
Does the marmotta config also need to use this variable, or are we punting that?

I suspect `360` might still be a bit conservative, but if you want to start there, that's fine with me.

markbreedlove:
I was thinking I'd make a second pass over the Marmotta and Nginx configurations, because there are settings in each that should be considered.

no-reply:
Sounds good.

PR #155, opened 2016-03-16
Add DPLA API key to primary-source-sets

markbreedlove:
See https://github.com/dpla/primary-source-sets/pull/126

This is necessary for the change in that PR to be deployed, but it's harmless for the config file to be modified without it.


AudreyAltman:
:+1:

PR #158, opened 2016-03-30
8329 contrib select

moltude:


moltude:
@markbreedlove Realized this was left off from last week and I forgot to create a PR. If you don't have time let me know, I just assigned you for review since the you wrote the original.

markbreedlove:
Thanks, @moltude. That looks good. I think you can remove https://github.com/dpla/automation/blob/8329-contrib-select/ansible/roles/exporter/files/export-provider.sh#L19-L25 if we don't need `$sed_opts` anymore.

anarchivist:
We do need `$sed_opts`, as it gets called by `outfile()`.

anarchivist:
Are we confident that the filenames generated will match the scheme that we have in place already? If so, then :rocket:

moltude:
@anarchivist let me take one more pass before merging

markbreedlove:
If I insert `echo` statements into the script in order to look at what would be generated, I see that it generates good filenames, as follows:
```
files$ ./export-provider.sh -u http://test:9200/test -o ~/tmp -s testbucket -p cdl
elasticdump --searchBody {
    "query": {
        "match_all": {},
        "filtered" : {
            "filter" : {
                "term" : {
                    "provider.@id" : "cdl"
                }
            }
        }
    }
} --input=http://test:9200/test --output=$ | gzip > /Users/mark/tmp/cdl.json.gz
aws s3 cp --quiet --content-type application/gzip --content-encoding identity /Users/mark/tmp/cdl.json.gz s3://testbucket/2016/03/cdl.json.gz
rm /Users/mark/tmp/cdl.json.gz
```

Does that look OK to everyone else?

moltude:
@markbreedlove it is the output filename that needs tweaking, currently it relies on the chaining of the two sed expressions and since it removed one of them it needed to be factored into this method 

https://github.com/dpla/automation/blob/8329-contrib-select/ansible/roles/exporter/files/export-provider.sh#L108

markbreedlove:
I may be overlooking something obvious, but the filename (cdl.json.gz in this case) is the same as it has been.

markbreedlove:
Ignore my comment above. @moltude have been chatting and I see what needs to be done.

markbreedlove:
@moltude: there are a couple of commits to fix upload file names, but I also mean to update the usage and comments to give good examples of how to specify the providers.

moltude:
@markbreedlove I'm testing the last two commit now and they look good

moltude:
@markbreedlove rebased, I think this is ready for release

PR #160, opened 2016-04-20
Add SSL capability and thumbnail proxy

anarchivist:
* Add thumbp role and playbooks.

* Configure site proxy NGINX for proxying thumbnail images.

* Remove X-Forwarded-Proto coercion to 'http' in frontend's NGINX
  config, which gets in the way of proxying SSL correctly.

* Change frontend settings.local to force 'https://' for API's base URL.
  FIXME: This is probably not good.  REVIEW BEFORE SUBMITTING PR.

* Configure frontend settings.local to use scheme-relative base URL for
  Wordpress and Wordpress widgets.

* Change Wordpress's wp-config.php to use SSL and scheme-relative URLs.
  Hardcode 'https://' for API URL.
  FIXME: This hardcoding is probably not good. REVIEW BEFORE SUBMITTING
  PR.  (It won't work with development, using the VMs, yet.)

### DIFF
--- a/ansible/group_vars/all.dist
+++ b/ansible/group_vars/all.dist
@@ -59,9 +59,12 @@ api_hostname: local.dp.la
 sitemap_host: sitemaps.example.tld
 marmotta_domain: ldp.local.dp.la
 
-# HTTP ports for various applications
+# HTTP ports/configuration for various applications
+#
 # api_port is the port the *loadbalancer* answers on for API requests
 api_port: 8080
+# api_default_protocol should not have trailing '://'
+api_default_protocol: http
 # api_app_port is the port the *backend app server* runs on
 api_app_port: 8003
 exhibitions_app_port: 8000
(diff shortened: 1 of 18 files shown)


anarchivist:
Ready for review; if it looks good, I'll test and squash.

ansible/group_vars/all.dist
```diff
@@ -59,9 +59,12 @@ api_hostname: local.dp.la
 sitemap_host: sitemaps.example.tld
 marmotta_domain: ldp.local.dp.la
 
-# HTTP ports for various applications
+# HTTP ports/configuration for various applications
+#
 # api_port is the port the *loadbalancer* answers on for API requests
 api_port: 8080
+# api_default_protocol should not have trailing '://'
+api_default_protocol: http
```
moltude: why was this reverted to http? These calls should be forced over https..

anarchivist: It wasn't reverted to `http`, it's just a reasonable default since the development environment isn't creating and deploying self-signed certificates. See dpla/aws#40.

moltude: got it, thanks

moltude:
@anarchivist I think this looks good and clear to merge

PR #165, opened 2016-05-03
Fix Marmotta installation

markbreedlove:
Do not use package from stack.linkeddata.org. Download and install the .war file from apache.org.

The current failure of apt commands related to the bad package repo is blocking various infrastructure tasks.

Procedure: (to be added to release notes):
After updating `automation`, run the marmotta role itself with one of the dedicated playbooks. That should fix apt.


moltude:
looks good

PR #166, opened 2016-05-03
Do not copy `.vagrant` directories

markbreedlove:
In roles that copy local development working copies of applications, exclude `.vagrant` directories.


markbreedlove:
Sorry to keep piling on the PRs, @moltude. I can reassign this if that helps. Hopefully this is a small one.

moltude:
👍

PR #167, opened 2016-05-04
add social media definitions to pss settings file

AudreyAltman:
This adds social media definitions to the primary source sets settings file.  They are used for various social media sharing features in the application.  These variables are used in other frontend applications, so the values for the variables are already defined.

I used this branch to run a test deployment of pss on staging and it worked as expected.

This impacts tickets [#8383](https://issues.dp.la/issues/8383) and [#8360](https://issues.dp.la/issues/8360).

### DIFF
--- a/ansible/roles/pss/templates/settings.local.yml.j2
+++ b/ansible/roles/pss/templates/settings.local.yml.j2
@@ -41,3 +41,12 @@ contact_email: {{ pss_contact_email }}
 
 action_mailer:
   delivery_method: {{ pss_delivery_method }}
+
+widgets:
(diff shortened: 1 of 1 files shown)


markbreedlove:
Without knowing how the pss application uses the variables, this looks good to me.

It does indeed appear that Tumblr still uses http, not https. If the Tumblr is not used as a base URL for assets, that should be OK.

AudreyAltman:
The Tumblr is used in the same way as it is in other frontend applications - it's just a link in the footer that takes you to the DPLA's Tumblr page.

PR #170, opened 2016-05-06
add api url to pss

AudreyAltman:
Add the API URL to primary source sets.  This relates to [PR#153 on PSS](https://github.com/dpla/primary-source-sets/pull/153) and [PR#42 on AWS](https://github.com/dpla/aws/pull/42).


moltude:
👍

PR #172, opened 2016-05-09
Change the `thumbp` role to install the Node.js rewrite

markbreedlove:
See https://github.com/dpla/thumbp/pull/4

I'll need to squash this before it's merged. "Fix ES port in init script" (499c732) should probably remain its own commit, but the others should probably be combined.


ansible/roles/thumbp/files/build_thumbp.sh
```diff
@@ -0,0 +1,33 @@
...
+cd /opt/thumbp/src/thumbp || exit_w_error "Can not cd to /opt/thumbp/src/thumbp"
+
+exit_w_error() {
+    echo >&2 $1
+    echo $1 >> $log
+    exit 1
+}
+
+echo "Ensuring correct version of Node" | tee -a $log
+nvm install || exit_w_error "nvm install failed"
```
moltude: @markbreedlove when I tested this locally it was failing at this step until I added $HOME/.nvmrc. Can you confirm?   Does it need to create a .nvmrc file as a part of install_nvm.sh?

markbreedlove: Can you confirm that you have `thumbp_branch_or_tag: fix/switch-to-node` in your `group_vars/development` file? That should have an `.nvmrc`.

moltude: that got it; thanks

moltude:
this looks 👍

markbreedlove:
OK, I've actually squashed it into three commits. I decided that the logrotate commit should be its own thing.

PR #173, opened 2016-05-10
Add apt-show-versions utility to common role

markbreedlove:
Add the apt-show-versions package to the list that is installed by the common role. This adds the command of the same name, which is a little easier to use than `aptitude` for determining which packages need to be upgraded.

Example usage:
```
$ ssh you@server apt-show-versions -u
```


moltude:
👍

PR #175, opened 2016-05-10
Fix/munin postgres plugin patch

moltude:
Addresses https://issues.dp.la/issues/8146 by adding tasks to find and replace Postgres 9.1 column names (current_query, procpid) with 9.4 column names (query, pid) in queries in the following munin plugins: 

- postgres_querylength_ 
- postgres_connections_
- postgres_connections_db
- postgres_users

Do not merge yet (need to squash)

### DIFF
--- a/ansible/roles/postgresql/tasks/main.yml
+++ b/ansible/roles/postgresql/tasks/main.yml
@@ -213,3 +213,40 @@
   tags:
     - munin
     - postgresql
+
+# Update queries in following munin plugins so they reference correct
+# PostgreSQL 9.4 column names
+#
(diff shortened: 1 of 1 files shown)


markbreedlove:
I'm observing this working in my development environment.

Are you OK with squashing the commits?

moltude:
thanks @markbreedlove -- I'll squash if it looks good to you.

markbreedlove:
I should note that I did not have to restart munin-node for the changes to take effect.

moltude:
@markbreedlove squashed

PR #176, opened 2016-05-12
Fix siteproxy wait_for_host port number

markbreedlove:
Fix siteproxy playbook's "Wait for site proxy node to come back up" task so that it uses the correct port number for the wait_for module. The port number is different in the development environment that it is
elsewhere.


moltude:
👍

PR #177, opened 2016-05-13
Redirect /wiki to Wayback Machine

markbreedlove:
Redirect /wiki to the Wayback Machine for historical Berkman DPLA wiki.


markbreedlove:
Assigning to @moltude but cc: @anarchivist

moltude:
look good. 👍

PR #179, opened 2016-05-20
Add default for pss_api_url variable

markbreedlove:
Add a role default for `pss_api_url`, which was added in https://github.com/dpla/automation/commit/47625481760b6237c8d8520288ba771d6db72e12.


markbreedlove:
This removes the necessity to modify a group_vars file to prevent an undefined variable error.

AudreyAltman:
👍

PR #180, opened 2016-05-20
Configure pss for memcache caching

markbreedlove:
This sets up the configuration necessary for the change introduced in https://github.com/dpla/primary-source-sets/pull/166


AudreyAltman:
👍

PR #181, opened 2016-05-20
Specify explicit PostgreSQL version in roles

markbreedlove:
In roles that install the postgresql-client package, specify the package name for the specific PostgreSQL version.  We use the apt.postgresql.org repository, which gives you version 9.5 if you specify the apt metapackage "postgresql-client".  We need version 9.4.

It would be nice to specify the `pg_version` constant in only one place, but there's no convenient place to do that outside of `vars/main.yml` in each relevant role.


moltude:
@markbreedlove do we need to create a task to uninstall the current version? If 9.5 was previously installed this doesn't seem to removed it and I assumed it would continue to be used if both postgres-client-9.4 and postgres-client-9.5 are installed.

markbreedlove:
I'll check to see what should be done about version 9.1, which we used to install, but we shouldn't have to uninstall version 9.5. The only reason 9.5 was on our own servers was because I was experimenting with it. I don't think I had `automation` set to install it, so it was just my own problem.

moltude:
Okay, it was showing up in my local VMs and it didn't seem to be cause any issues. I didn't check the production or staging postgres envs.  

I think this should be 👍 if there isn't a problem w/other versions

markbreedlove:
In that case, if you had it in your local VMs, then I may be forgetting something. I'll check the `automation` commit history to see if I had it installing in the past.

markbreedlove:
I've made an oversight, and will close this PR.

It is OK for the postgresql-client to be at version 9.5, according to the `pg_wrapper` script, which `/usr/bin/psql` symlinks to: https://alioth.debian.org/scm/loggerhead/pkg-postgresql/postgresql-common/trunk/revision/1216

Please ignore my confusion about uninstalling 9.1 or 9.5. The presence of the client package is unrelated to my experimentation in development with the 9.5 server package.

PR #182, opened 2016-05-23
Improve NGINX cache configuration

markbreedlove:
The goal of these changes was originally to cache `/thumb` resources, but it became apparent that we could do better in caching overall.

With this commit, and some work on the frontend app, it will be possible some day to forego the cache-purging that we usually do after a deployment.

* Reduce the number of proxy_cache_valid directives.

* Express 'expires' with date units.

* Remove redundant proxy_cache_valid directives. There was some variation, but not enough for it to be perpetuated.

* Rename shared memory zone from "staticfilecache" to "main_cache" because it's not simply a static file cache.

* Add "inactive" parameter to proxy_cache_path, which was defaulting to 10 minutes.

* Add note about proxy_cache_revalidate (NGINX sends If-Modified-Since to origin).  We can't use it because we're still on Ubuntu 12.04LTS, with NGINX 1.1.

* Add proxy_cache_lock so concurrent requests don't get sent to origin server for the same thing.

* Change some comments.

* Change proxy_cache_use_stale for /thumbp location. Move on to next origin server in the event of connection error or timeout via proxy_next_upstream, but deliver stale file for a 500 or 503, assuming the other server won't respond any better.

* Add Cache-Control and X-Cache-Status headers to individual locations

* Hide X-Powered-By header.

* Shorten expiration time of Exhibitions assets.

* Enable caching of Primary Source Sets assets.

* Have the cache store 301 responses in a number of locations.

* Remove test.php and phpinfo.php locations within /info location.

* Make proxy_cache_key consistent across configuration.

* Remove ability to defeat caching by setting Cache-Bypass header.

* Simplify use of proxy_cache_bypass and proxy_no_cache.

It would be great to extend the expiration of frontend assets, but there are still some that the `frontend` app does not seem to compile and assign cache-extending names. The expiration for assets in the frontend therefore is one minute.

Wordpress assets in wp-includes and in plugins are versioned, but those in the berkman theme are not. No image assets are versioned. The expiration for the wordpress site for anonymous users (who do not have the relevant cookies) is one minute for most resources and one month for the versioned assets.

These changes don't fix the development environment's longstanding asset-pipeline problems. The frontend has uncompiled assets that shouldn't be cached for long periods. The deployments still clear the proxy cache, so this is OK for the meantime. In the future, we need to clear the cache conditionally: clear it after a deployment only in the development environment, and let asset versioning do the job in production.

The `_dpla_no_cache` cookie for the frontend should still be recognized because it still applies to admin requests. For caching to be optimal, the frontend app needs to be modified to set the `_dpla_no_cache` header only when necessary, not with every single request. It should only be set when an admin logs in or out. The `/saved` and `/users` locations have caching disabled for regular users who are logged in.

We don't cache the frontend app's HTML because it's too problematic for loggged-in users and the app library. Applications should be in charge of setting good cache headers for authenticated and unauthenticated requests, and it's bad to second-guess them in the proxy configuration.

### DIFF
--- a/ansible/roles/site_proxy/defaults/main.yml
+++ b/ansible/roles/site_proxy/defaults/main.yml
@@ -42,3 +42,7 @@ siteproxy_nginx_limit_req_log_level: notice
 siteproxy_nginx_cache_max_size: 6144m
 # Size of shared memory zone that is used for NGINX cache
 siteproxy_nginx_cache_shmem_size: 360m
+# How long an item can remain in the cache without being accessed, independently
+# of how long its expiration time is, or the value of the proxy_cache_valid
+# directive:
+siteproxy_nginx_cache_inactive_time: 1d
--- a/ansible/roles/site_proxy/templates/etc_nginx_sites_available_site-proxy.j2
+++ b/ansible/roles/site_proxy/templates/etc_nginx_sites_available_site-proxy.j2
@@ -1,21 +1,33 @@
 
 # proxy configs
-proxy_cache_path  /var/cache/nginx levels=1:2 keys_zone=staticfilecache:{{ siteproxy_nginx_cache_shmem_size }} max_size={{ siteproxy_nginx_cache_max_size }};
+proxy_cache_path  /var/cache/nginx
+                  levels=1:2
+                  keys_zone=main_cache:{{ siteproxy_nginx_cache_shmem_size }}
+                  max_size={{ siteproxy_nginx_cache_max_size }}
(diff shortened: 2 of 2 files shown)


markbreedlove:
Related ticket: https://issues.dp.la/issues/8348

no-reply:
Looks good. I was able to confirm caching on my local VMs.

Thanks for the detailed commit message. It's a big help in reviewing something like this. :rocket:

PR #184, opened 2016-05-31
Redirect HTTP to HTTPS

markbreedlove:
Make sure that all frontend and CMS sites are HTTPS.

Simplify redirect for "www" to use `return` instead of `rewrite`, while adding necessary `default_http_scheme`.


no-reply:
:+1:

moltude:
👍 👍

PR #185, opened 2016-06-07
Facilitate dynamic inventory and custom environments

markbreedlove:
* Add elasticsearch_index role. This will be incorporated into the development playbooks at a later point.
* Update some variables to default to False when they're not defined.
* Parameterize some configuration file parameters.
* Change common role to observe `all_group_name` for more flexibility with custom environments.
* Update common role to work better in dynamic inventory situations.
* Update postgresql role to use IP addresses instead of hostnames, for simplicity, and to avoid having to convert characters in dynamic-inventory hostnames.
* Tidy up various minor things.


no-reply:
:+1:

PR #187, opened 2016-07-01
Stop clearing NGINX proxy cache unnecessarily

markbreedlove:
The NGINX cache configuration was changed in [v3.3.0](https://github.com/dpla/automation/releases/tag/v3.3.0) to make it safer for the cache to hold on to assets for a longer time. Pages that have to be fresh are no longer cached. We can therefore remove the role tasks that used to clear the cache. We still clear the Memcache cache where necessary, and we still have the playbook for doing it manually.

Addresses https://issues.dp.la/issues/8157

This could be something to try with one of your next deployments to verify how it goes.

### DIFF
--- a/ansible/roles/api/tasks/deploy.yml
+++ b/ansible/roles/api/tasks/deploy.yml
@@ -74,10 +74,5 @@
   notify:
     - restart memcached
 
-- name: Clear site proxy cache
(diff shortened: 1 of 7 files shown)


anarchivist:
Looks good. 👍

moltude:
affirming 👍

PR #189, opened 2016-07-14
Install geocoder config for ingestion

markbreedlove:
Install the geocoder.yml configuration file that is needed by the Audumbla gem for enrichments in the ingestion application.

The installed configuration will not work in a development environment until the developer sets the `ingestion_app_geo_host` variable to something suitable (for example, in the `ansible/group_vars/development` file). The Twofishes geocoding server is not installed into the development VMs and must be managed separately.

Fixes https://issues.dp.la/issues/8455


no-reply:
:+1: Thanks @markbreedlove!

anarchivist:
👍

PR #190, opened 2016-08-02
Check for errors in more build script locations

markbreedlove:
Add checks for return statuses in more of the commands issued in Rails deployment scripts.  There were rake tasks and other commands that could have failed silently and could have allowed deployments to proceed incorrectly.


AudreyAltman:
This looks good to me.  Thanks @markbreedlove!  Do you want anyone else to look at it before I merge?

no-reply:
:+1:

PR #191, opened 2016-08-03
Fix thumbp pidfile location

markbreedlove:
Instead of creating a subdirectory of `/var/run` that will get wiped out upon reboot, just put the pidfile in `/var/run` as other programs do.


no-reply:
:+1:

anarchivist:
👌

PR #192, opened 2016-08-22
Add ec2_swapper role

markbreedlove:
This role is for EC2 instances that need to have swap partitions.

It is not used by any of the development inventories that ship with `automation`.

This method of creating a file with `fallocate` and mounting it as swap gives us the most consistency between different types of EC2 instances and will make automated provisioning easier.


no-reply:
:+1:

PR #193, opened 2016-08-25
Add missing do_deployment var to thumbp playbook

markbreedlove:
Add the `do_deployment` variable to the `thumbp.yml` playbook.

This variable is evaluated by a number of roles, where it has no default by design. It was missing from the `thumbp` playbook, causing an error.


no-reply:
:+1:

PR #194, opened 2016-08-26
Facilitate sandbox or demo sites

markbreedlove:
- Add ability for test, demo, or development hosts to serve SSL without a loadbalander by adding parameters to the `site_proxy` role.

- Change `site_proxy` NGINX configuration to leave out location blocks for sites that are not enabled.  Additionally, allow for demo sites where there is no frontend app, and there needs to be a redirect from `/` to the app that is running.

- Move self-signed certificate generation script from `dev_loadbalancer` role, modifying it for more general usage in more settings.


anarchivist:
Looks OK to me. Needs to be rebased.

markbreedlove:
There's the rebase.

PR #196, opened 2016-09-20
Add open files ulimit settings for Marmotta Tomcat

markbreedlove:
Ensure that Marmotta's Tomcat container process has a high-enough open-files limit by adding /etc/security/limits.d/tomcat.conf with settings for the `tomcat7` user.

The new number, 10240 is 10 times the limit of 1024 that was being set in our default Ubuntu 14.04 install.


markbreedlove:
I _was_ actually thinking about that last night when I was offline.
I will do that.  Could be useful for different scenarios to parameterize the open files.

markbreedlove:
@anarchivist I've templatized the file

anarchivist:
Looks good to me. Anyone else want to take a look before :merge:?

PR #198, opened 2016-09-21
Move Tomcat kernel limits task to tomcat_common role

markbreedlove:
Move the task and variables to set kernel limits for the Tomcat process out of the `marmotta` role and into the `tomcat_common` role.

To set specific limits, e.g. for Marmotta or another Java application, one should modify the `host_vars` file for the relevant server.


markbreedlove:
@moltude @mdellabitta bump

moltude:
👍  merging

PR #199, opened 2016-09-29
Configure logfile rotation in common_web role

markbreedlove:
Add a task to configure logrotate for the NGINX logs, with a variable to govern the rotation period.

Remove the logfile rotation task and template from the Marmotta role.

Note that if different hosts need specific log rotation periods, they can be configured in host_vars files.


markbreedlove:
@moltude @mdellabitta bump

moltude:
this looks 👍

PR #201, opened 2016-10-12
Add more kernel param vars to dev2 in inventory

markbreedlove:
I've had this change sitting around for a while without committing it. It has worked fine with my VMs (where I had them in a `host_vars` file, but the effect should be the same).


markbreedlove:
I've updated the summary above. These parameters are related to PostgreSQL and I should have committed them for `dev2` in the past.

PR #202, opened 2016-10-12
added location alias for siteproxy.xml

mdellabitta:
Just a quick addition to the nginx config to make the root sitemap available at the root of our site for testing Google Webmaster Tools against.


markbreedlove:
Would it help if I added a task to the `site_proxy` role to copy this file into place, perhaps retrieving it from a remote location?

markbreedlove:
Hi, @mdellabitta. Do you think this is safe to close now, since your discovery of the existing sitemap support? Are the S3 files working?

moltude:
Bumping open question to @mdellabitta. Should this be closed out since it appears that the current site maps are working?

mdellabitta:
Yeah I believe we can clear this because the mechanism for surfacing sitemaps works in a different way.

PR #203, opened 2016-10-19
Revive legacy contentqa engine

markbreedlove:
Revive the deployment and configuration of the legacy `contentqa` rails engine, inside the `platform` app.  Also improve some of the configuration that used to be there.

The `contentqa` engine is irrelevant to most installations and is disabled by default.


ansible/playbooks/dev_loadbalancer.yml
```diff
@@ -7,4 +7,4 @@
   hosts: loadbalancer
   become: yes
   roles:
-    - { role: dev_loadbalancer, tags: ['dev_loadbalancer', 'loadbalancer'] }
+    - { role: standalone_loadbalancer, tags: ['standalone_loadbalancer', 'loadbalancer'] }
```
moltude: Why/should this file be renamed to `standalone_loadbalancer.yml` to keep naming conventions?

moltude: Just to follow-up after the ddos nonsense on Friday. I'm not referring to the role label but the file itself (which is still named `dev_loadbalancer.yml`. I agree with the changes `standalone` elsewhere but was suggesting that this file be renamed as well. @mb is that clear? Sorry if we were talking past each other

markbreedlove: Sure, that's clear! Thanks, and yes, makes sense to rename the file, too.

markbreedlove: In reviewing how this particular playbook is used, I need to reverse what I said above. It actually makes sense for it to be `dev_loadbalancer` because this playbook is only used in the context of the development VMs. Though the role has been generalized, and we use it with an internal playbook that we don't share, we don't include this playbook in that internal one, and shouldn't have to.

I think that I will change the name of the role, though, to "loadbalancer."

markbreedlove:
@moltude: I have changed the role name to `loadbalancer` in https://github.com/dpla/automation/pull/203/commits/24e535462698f4d373b7789ad72807f926b2e2cd but have left the name of `dev_loadbalancer.yml` alone, per https://github.com/dpla/automation/pull/203#discussion_r84755412

moltude:
This looks 👍 to me and I'm ready to merge it in

PR #204, opened 2016-11-02
Lower defult NGINX cache size

markbreedlove:
Lower the default size of the NGINX disk cache for the site_proxy role from 6Gb to 5Gb. This role is likely to be deployed on AWS, where the default volume size is 8Gb, so 6Gb is not leaving a lot of room for other things on the filesystem.


moltude:
👍

PR #207, opened 2016-11-02
remove un-needed setting for frontend exibitions elasticsearch

AudreyAltman:
Remove un-needed frontend setting related to searching for exhibitions in elasticsearch.

This addresses [ticket #7579](https://issues.dp.la/issues/7579) and relates to [frontend PR#131](https://github.com/dpla/frontend/pull/131).


markbreedlove:
I have to assume that a `frontend` installation would break if using this commit and using an old release of `frontend` that still had the Exhibitions search. I'm pondering how to handle that. We've removed features before from apps and I need to think how this relates.

markbreedlove:
OK, thanks!

PR #211, opened 2016-12-02
Fix mount / unmount in ec2_swapper role

markbreedlove:
Fix the usage of the `mount` module in the `ec2_swapper` role so that it removes Ubuntu's `fstab` entry for `/mnt` when it unmounts it.  (The Ansible documentation is actually wrong about the difference between "unmounted" and "absent!")

Also remove the unnecessary `file` task to create the `/swap` directory, because the `mount` module will create it lower down in the file with `state=mounted`.


moltude:
👍  too

PR #213, opened 2016-12-02
Improve and fix two top-level playbooks

markbreedlove:
Improve staging_all.yml and production_all.yml.  Add role-level tags and add an empty play to gather facts for all hosts as insurance against undefined variables.


mdellabitta:
Oh wait, looks like a colon is missing...

markbreedlove:
yikes. hm. I'll fix it

PR #215, opened 2016-12-02
Add new variable for all-in-one hosts

markbreedlove:
Add a new variable, `all_in_one_hosts`, which governs whether each host in the inventory is self-contained, with no dependencies on other servers. This defaults to False, so no changes should be necessary for typical developer usage. This is for particular DPLA-specific use cases, like QA, and is not relevant to either production or development VM environments.

Also normalize some of the dict variable dereferencing syntax throughout various files, for a cleaner look, and for consistency. For example, change `groups['cms']` to `groups.cms`.

### DIFF
--- a/ansible/group_vars/all.dist
+++ b/ansible/group_vars/all.dist
@@ -79,6 +79,11 @@ thumbp_port: 8007
 nginx_real_ip_from_addrs:
     - 192.168.50.0/24
 
+# Whether each host has all of the services in a local silo, as would be the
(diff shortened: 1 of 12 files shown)


markbreedlove:
Though this shouldn't be disruptive to _us_ at DPLA, it could be to someone else who had some kind of all-in-one server that used the variables that this changeset has superseded. We'll have to think of that when cutting a release.

moltude:
👍  but holding merge incase @mdellabitta wants to review

PR #223, opened 2016-12-20
Add assertion of libssl-dev package, for frontend

markbreedlove:
This package is required for `rbenv` to compile Ruby with SSL support, which is necessary when running `bundle install`. There seems to have been a recent change (within the last week) in an Ubuntu package's dependencies that has stopped `libssl-dev` from being installed without our explicitly requesting it.


moltude:
👍

PR #224, opened 2017-03-17
Upgrade Ubuntu to 14.04 LTS

markbreedlove:
Update README files with updated installation instructions.

Update Vagrantfiles with new vagrant box names, and more memory allocation for the 'loadbal' VM.

Update 'common' role with changes to configuration files that coordinate with new package versions.

Update 'dbnode' role to work around broken dependency in 'bigcouch' package that was intended for Ubuntu 12.04.

Update 'pss' role to install libssl-dev dependency that is required by rbenv's build of Ruby 2.1. There must have been a depcndency cascade in Ubuntu 12.04 that caused it to be installed previously, whereas it was missing under 14.04.

It seems like there is more memory being allocated on these VMs with Ubuntu 14.04, which can lead to out-of-memory situations; so reduce memory usage by the following means:
* Lower the default PostgreSQL max_connections and shared_buffers
* Lower the default NGINX cache maximum size and shared memory size
* Lower the default number of Unicorn workers across various roles

Also allocate more memory to webapp1 to get around out-of-memory problems with the frontend build.

### DIFF
--- a/README-ingestion2.md
+++ b/README-ingestion2.md
@@ -15,12 +15,11 @@ See the upgrade notes in [README.md](README.md).
 
 ## Steps
 
-* Copy `Vagrantfile.ingestion2` to `Vagrantfile`
+* Copy `Vagrantfile.ingestion2` to `Vagrantfile`.  (Not `Vagrantfile.dist`)
 
-* Make sure that Vagrant has downloaded the base server images that we'll need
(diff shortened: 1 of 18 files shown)


mdellabitta:
This has been tested thoroughly on staging and in vagrant environments and seems to work well. 👍

PR #225, opened 2017-03-20
Improve "clean" deployments, various apps

markbreedlove:
Fix an oversight in the `thumbp` role, where a re-created `thumbp` user account would result in `/var/log/thumbp/thumbp.log` no longer being writable after a "clean" deployment, which would cause `thumbp` not to be able to start again.  (Solution: recursively ensure log directory ownership.)

Introduce the `clean_rbenv` variable, which can be used to delete a Rails application's Rbenv in order to clean it out, or to recompile C extensions from scratch; useful, for example, after an o/s upgrade.

### DIFF
--- a/ansible/roles/api/tasks/deploy.yml
+++ b/ansible/roles/api/tasks/deploy.yml
@@ -54,6 +54,14 @@
       src=unicorn.rb.j2 dest=/home/api/api/config/unicorn.rb
       owner=api group=api mode=0644
 
+# To use "Clean rbenv", define the variable `clean_rbenv', which is not defined
+# anywhere by default.
+- name: Clean rbenv
+  file:
+    path: /home/api/.rbenv
+    state: absent
+  when: clean_rbenv | default(false)
+
 - name: Make sure rbenv and bundler are current
   script: >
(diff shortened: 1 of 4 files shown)


moltude:
Should the cleaning of rbenv apply to PSS as well?

markbreedlove:
Thanks, @moltude, you've pointed out an important flaw in all of this. We should either have separate application user accounts for `pss` and the others, whereas we now have just the `dpla` account; or the clean operation should just remove the relevant rbenv version. Otherwise, on a development VM, this could wipe out the rbenv of, say, the frontend, when pss is cleared, and vice-versa.

I'll make an issue tracker story to follow up on this work next sprint.
Thanks again for looking at this.

moltude:
Sure thing @markbreedlove I understand your comment to mean that you'd like to keep this PR open for that new ticket. Is that correct?

markbreedlove:
Thanks, that's correct. It should be closed out as part of the scope of the next ticket.

markbreedlove:
@moltude: I've finally added PSS, since those other account-related changes have gone through. I'ver rebased this on `develop`.

PR #226, opened 2017-03-22
Add template for debugging variables

markbreedlove:
Add templates/variables.j2, with inline usage instructions.

Reviewers should be able to try this out on a vm or wherever.


AudreyAltman:
👍

PR #227, opened 2017-03-28
Fix failure of php-fpm to be respawned by upstart

markbreedlove:
The upstart (init) configuration file for the php5-fpm package in Ubuntu 14 fails to respawn the php-fpm master process when it's killed with SIGKILL or (presumably) segfaults or dies due to some other reason.

For our purposes, we only want a SIGTERM or maybe SIGINT to be considered an intentional stop signal that signifies success. We also want to make sure that stale child processes are killed off before trying to restart php-fpm, because they'll be holding on to the socket files, preventing the service from starting.

This fixes https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1230


mdellabitta:
This looks good to me.

moltude:
Looks good to me too.

PR #228, opened 2017-04-05
Add Universal Viewer

markbreedlove:
Add `uv` role for Universal Viewer, and update NGINX configuration to serve the Javascript under `/uv`.

This changeset interacts with ongoing changes being made in the `frontend` app, so it shouldn't be merged until everything is synced up between `automation` and `frontend`.

To test, try running the role in your development VMs. You can run `ansible-playbook` like this:
```
ansible-playbook -u <you> -i development playbooks/dev_webapps.yml -e 'level=development'
```
If you need to deploy it again from scratch you can run the same playbook with `-e 'level=development clean_uv=true' -t uv` and save some time.

Let me know if you get an error from the `npm install` step in `build_uv.sh` saying "Fatal error in ../deps/v8/src/heap/store-buffer.cc" and "CHECK(object->IsHeapObject()) Failed".  This may be an out-of-memory issue that I need to fix. It's been intermittent for me; I've always found that running `npm install` again works when it fails the first time. This is something that needs to be resolved before this is merged, but I want to get this out in front of you in the meantime to work out the necessary integration with `frontend`.

### DIFF
--- a/ansible/development
+++ b/ansible/development
@@ -44,6 +44,9 @@ dbnodes
 [thumbp:children]
 webapps
 
+[uv:children]
+webapps
+
 [development]
 loadbal
 dbnode1
--- a/ansible/playbooks/dev_webapps.yml
+++ b/ansible/playbooks/dev_webapps.yml
@@ -41,6 +41,7 @@
   roles:
     - frontend
     - { role: thumbp, tags: ['thumbp']}
+    - { role: uv, tags: ['uv']}
   vars:
     level: development
     do_deployment: true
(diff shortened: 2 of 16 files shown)


markbreedlove:
This will need to be squashed before being merged.

markbreedlove:
Regarding the comment in the PR description, the work on `frontend` is done for now, so this should be OK to consider merging.

mdellabitta:
Tried out an install and it does what it says on the tin.

PR #231, opened 2017-05-05
Remove "bad bot" exclusion, site proxy Nginx config

markbreedlove:
Remove ineffective attempt at excluding bad bots in the site_proxy NGINX configuration. It blocked a seemingly random and very small number of bots by user-agent. Using this methodology it would be impossible to keep the list fresh, and, indeed, it's never been maintained. Besides, it assumes that "bad bots" are identifying themselves honestly.


markbreedlove:
bump :-)

PR #232, opened 2017-05-10
Improve headers related to proxying

markbreedlove:
This cleans up a number of NGINX-related files, and should make HTTP transactions more consistent between development and production. It allows more flexibility in using different kinds of loadbalancers in different configurations, as long as they pass `X-Forwarded-For`, which we should be able to expect.

Additionally, this hides the `X-Powered-By` headers that are set by some web application frameworks, where they weren't being hidden already.


mdellabitta:
I've pointed out one minor thing via separate cover which may not matter, but otherwise, I approve!

markbreedlove:
Because of that last "fixup" commit I should squash this before it's merged.

mdellabitta:
GFI, then I'll merge.

markbreedlove:
OK, should be clear now.

PR #233, opened 2017-05-15
Make separate accounts for applications

markbreedlove:
Make separate o/s accounts for individual web applications, instead of having all of them use "dpla".

This improves our ability to keep deployments from interfering with one-another and to do "clean" deployments.


markbreedlove:
See https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1240

PR #236, opened 2017-06-07
Move Ingestion 1 config to a role

markbreedlove:
Remove the ingestion.yml playbook and add a role for setting up Ingestion 1.

This does not apply the role to the development VMs because we still need to evaluate whether webapp1 has enough memory available for that.

This makes it easier to make custom stacks (like we do at DPLA for content QA) that are prepped for installing https://github.com/dpla/ingestion


moltude:
@markbreedlove Why was the owner:group of /v1/ingestion changed to `ingestion` from `dpla`?

markbreedlove:
@moltude Because account names always should have been more descriptive and we shouldn't have the omnibus accounts that serve multiple applications. Also just for consistency because we've needed to swap out the `dpla` account in other places. If you want to start from scratch with one application you don't want to blow away environment and configuration settings for the others.

PR #238, opened 2017-07-17
Add jenkins

markbreedlove:
Add a `jenkins` role to configure a Jenkins master server.

Add a generalized `pyenv` installer.

Additionally (as a separate commit), add a condition to decide whether to copy the SSH key for deployment of the private assets deployed with the Primary Source Sets application.

### DIFF
--- a/ansible/roles/jenkins/tasks/main.yml
+++ b/ansible/roles/jenkins/tasks/main.yml
@@ -1,0 +1,149 @@
+---
+
+- name: Ensure Java 8 runtime and 'daemon' installations
+  # Java 8 is available on Ubuntu 16.04 LTS (xenial), not 14.04 LTS (trusty)
+  apt:
+    pkg: "{{ item }}"
+    state: present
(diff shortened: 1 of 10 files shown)


ansible/roles/jenkins/tasks/main.yml
```diff
@@ -0,0 +1,149 @@
+---
+
+- name: Ensure Java 8 runtime and 'daemon' installations
+  # Java 8 is available on Ubuntu 16.04 LTS (xenial), not 14.04 LTS (trusty)
+  apt:
+    pkg: "{{ item }}"
+    state: present
+    update_cache: yes
+  with_items:
+    - openjdk-8-jre
```
mdellabitta: I'd probably default to jdk instead of jre if only because you get some of the diagnostic cli tools like `jstack`.

markbreedlove:
I'll try changing the package to JDK before this is merged ...

PR #239, opened 2017-09-01
Fix ec2_swapper role

markbreedlove:
* Do not use instance store volumes at all. They go away if you stop and start an instance. (As distinct from restarting.)

* Add playbook

### DIFF
--- a/ansible/playbooks/swapper.yml
+++ b/ansible/playbooks/swapper.yml
@@ -1,0 +1,9 @@
+---
+
+
+- hosts: all
+  become: yes
+  gather_facts: yes
+  roles:
+    - {role: ec2_swapper, tags: ['ec2_swapper']}
+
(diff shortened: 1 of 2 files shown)


mdellabitta:
So if I understand this correctly, this will end up making swap files in /swap, most likely on the root volume?

markbreedlove:
Correct, on the root volume, which is always on an SSD drive. So far this has not incurred any performance hit.

mdellabitta:
Seems safer, too. I assume swap is mostly there to prevent important processes from being reaped.

markbreedlove:
Yes, there's less to think about this way. It's there in case it's needed to keep a process from running out of memory, or if the kernel wants to proactively tuck some little-used pages out of the way for better caching, but we don't expect it to be used often.

mdellabitta:
👍

PR #242, opened 2017-09-22
Jenkins ingestion3

markbreedlove:
Improve the `jenkins` role. This is for managing the state of a Jenkins master.

Add the `s3cmd` utility (a Python module) which is useful at least for ingestion3-related file transfers, if not for more things in the future.

Assuming we only have to worry about one version of Python relative to Jenkins's scripts, just set the `pyenv` python version globally.

Also change the global node version manager (nvm) installer script (used by other roles in `automation`) to write a per-user  log file so that multiple per-user installs on one system are possible without file permissions errors.


ansible/roles/jenkins/defaults/main.yml
```diff
@@ -11,3 +11,5 @@ jenkins_sha256: 34fde424dde0e050738f5ad1e316d54f741c237bd380bd663a07f96147bb1390
 jenkins_md5: 9e14790b2345bf673abb24224eee4d2c
 jenkins_python_version: 2.7.13
 jenkins_aws_region: us-east-1
+jenkins_aws_key: override_me
```
mdellabitta: Maybe I'm not understanding what this does, but shouldn't we rely on IAM injecting this? 

I believe it already does by virtue of being granted the full access role for ElasticBeanstalk. If not, we can grant access to the S3FullAccess policy...

markbreedlove: Let me follow up on that. The original reason I added this was because the Spark library did not recognize the instance key.  i.e. using `s3://` urls as paramaters to `MappingEntry`.  Then, when I was setting up s3cmd it seemed to want those values in its config file. I can do an experiment by taking out those values from the s3cmd config to see if the E2 instance key takes over.

markbreedlove: oh, but another thing is if you want to develop on a local vm wihtout messing with the production box it's not going to be an ec2 instance that's set up with an ec2 key.

mdellabitta: Ideally in production, we can make use of the IAM injected key in prod, because that auto rotates the key every hour or something, which is the main reason I brought it up.

mdellabitta: I believe a new enough version of s3cmd should work with the auto rotated keys in the ec2 instance metadata. I believe I've seen that working.

markbreedlove: OK, it does indeed work on EC2 if you get rid of the key parameters from the `.s3cmd` file, meaning it falls through to the EC2 instance IAM credentials.

For development work, I will add an `environment{}` section in the Jenkinsfile to set the AWS credentials.

I'll remove the variables you noted and adjust the s3cmd configuration template, and submit another commit.

markbreedlove: This does mean that if we fix the problem with our Entry objects' `s3://` parameters (allowing you to pass S3 urls once again) we'll have to readdress this.

mdellabitta: If we have a new enough AWS jar, I'm pretty sure it knows how to pick up this from instance metadata, too.

PR #250, opened 2018-03-20
Improve PSS deployment with local source

markbreedlove:
Improve the PSS deployment using the "local source" option so that it's easier to simulate the production environment with the development VMs.

The old method, copying the configuration files from the developer's source code working copy, was confusing, because they're normally written from the .j2 templates. If you want to use "local source" and also have config files that are similar to what you'd get in production, it's easier to manage that by updating your `group_vars/development` file with the settings you want.


markbreedlove:
In case that's not clear, it means you can have something like this in your "development" vars file:
```
pss_rails_env: production
pss_secret_key_base: fef68[...]6578
pss_use_local_source: true
```

PR #256, opened 2018-07-31
Upgrade Elasticsearch to version 6

markbreedlove:
* Update 'elasticsearch' role to use Elasticsearch 6.
* Update elasticsearch-related playbooks. Remove initial_run variable,
  because shard replication issues will be addressed with changes to
  each index's settings.
* Remove unneeded files from role
* Update README.md with different hosts
* Update Vagrantfile.dist and 'development' inventory with different VMs
* Fix `roles/common/tasks/main.yml` to make sure that the 'admin' group
  exists on Ubuntu 16, which is what we use for Elasticsearch.
* Add elastichq role, for monitoring tool ElasticHQ.
* Remove obsolete delete_undeployed_indices.sh
* Fix `common_*` playbooks for Ubuntu 16
* Ensure that Python is found on Ubuntu 16


ansible/roles/elasticsearch/templates/elasticsearch.yml.j2
```diff
@@ -1,554 +1,111 @@
...
+
+# ----------------------------------- Memory -----------------------------------
+#
+# Lock the memory on startup:
+#
+bootstrap.memory_lock: true
+
+
+#
+# Make sure that the heap size is set to about half the memory available
```
mdellabitta: :thinking:

markbreedlove: Yeah, I don't get that exactly either. The heap size is about 26% of RAM right now on our cluster. This comment was unchanged from what I downloaded.

ansible/roles/elasticsearch/templates/jvm.options.j2
```diff
@@ -0,0 +1,103 @@
+## JVM configuration
```
mdellabitta: Some good stuff in here.

PR #259, opened 2018-10-19
Remove API, BigCouch, "exporter," and "elasticsearch_index" roles

markbreedlove:
The `api` and `dbnode` roles are obsolete now with the rollout of the new API application.

The `elasticsearch_index` role became obsolete when we converted to Elasticsearch 6, and since we're now using `sparkindexer`.

The `exporter` role is obsolete, because we don't produce its "bulk downloads" files anymore. Our documentation now refers people to our S3 master dataset bucket.


mdellabitta (approved):
We do still point people at the exports rather than the master dataset, but it's fine to remove the exporter. it doesn't seem to be very reliable. We're looking at doing an export with spark that will support more formats.

mdellabitta (approved):

